PR #3, opened 2018-05-15
Stub out the FastCGI API

pushkarnk:
This code is copied from https://github.com/IBM-Swift/Kitura-net/tree/master/Sources/KituraNet/FastCGI

The stubs need to be replaced by proper SwiftNIO based implementations.


pushkarnk:
Verified by building locally

PR #13, opened 2018-05-24
Fix SSL test failures on Darwin

nethraravindran:
`swift nio-ssl` doesn't add support for PKCS#12-formatted certificates. Hence we test only with PEM formatted certificates.


pushkarnk:
All tests passed after a couple of restarts. We need to understand the random failures better.

Merging.

PR #37, opened 2018-06-27
Support Kitura-WebSocket-NIO - internal API changes

nethraravindran:


Sources/KituraNIO/HTTP/HTTPServer.swift
```diff
@@ -37,6 +37,8 @@ public class HTTPServer : Server {
 
     private let syncQ = DispatchQueue(label: "HTTPServer.syncQ")
 
+    private var ctx: ChannelHandlerContext?
```
pushkarnk: @nethraravindran The HTTPServer instance never goes out of scope. Wondering if this would end up holding the related `ByteBuffer` instance forever.

pushkarnk:
Thanks @nethraravindran . Merging.

PR #38, opened 2018-06-27
setting authorization header if username and password are not nil

saiHemak:
Problem: Authorization headers are not getting set by Kitura-NIO to the request header even if username and password  fields are set by the client.

Authorization header needs to be set based on the auth-scheme  preferred by the user.If username and password are set, treat it as basic authentication scheme and set the authorization header accordingly.


pushkarnk:
LGTM

PR #43, opened 2018-07-03
Prepare the Date header value periodically

pushkarnk:


pushkarnk:
Neither the HelloWorld benchmark nor TechEmpower seem to be indicating any performance gains with this change. Hence closing.

PR #50, opened 2018-07-18
Increase the WebSocket frame size

nethraravindran:
Websocket upgrade fails for a larger payload above 16,384 bytes. This is because the default frame size for the websocket decoder  to tolerate is only 2^14 (16,384) bytes. Hence we can increase the frame size for the decoder to tolerate upto 2^16 ie. 65,536 bytes.


pushkarnk:
From the discussion with @nethraravindran , it is apparent that if a payload with length greater than the `maxFrameSize` is received, we abort the protocol with the 1009 error code. This means there is no other mechanism that the protocol implementation falls back to, when the payload size exceeds our limit. I am hence okay with increasing the limit from 2^14 to 2^16 for the sake of Autobahn testing. However it must be noted that the larger this limit, the more vulnerable the implementation becomes for denial of service attacks.

Issue #63, opened 2018-08-16
ThinkTime test route /think/async fails with KituraNIO


nethraravindran:
ThinkTime test with the route `/think/async` crashes in Kitura-NIO. 

The crash is because the thread that initiated the I/O task is not the same thread that performs the I/O operation.  ie, the current thread that tries to perform the operation is not as the same thread that was tied with EventLoop. 

Hence we must ensure that we perform the I/O task on the same thread or we submit the task to be executed to the EventLoop.

nethraravindran:
PR : https://github.com/IBM-Swift/Kitura-NIO/pull/64

pushkarnk:
@nethraravindran Can you provide a description of the bug please?

pushkarnk:
#64 merged. Thanks @nethraravindran

PR #64, opened 2018-08-16
Submit and execute I/O on EventLoop

nethraravindran:


pushkarnk:
@nethraravindran Can we have a commit message indicating what this change does?

pushkarnk:
Fixes #63

PR #68, opened 2018-08-17
Check if the URI is valid before upgrading

pushkarnk:
For an upgrade to a protocol like WebSocket, we must be able to check if the requested URI has a service registered. If not, `shouldUpgrade` must flag that error by returning nil. This will cause the NIOWebSocket upgrader to throw `unsupportedWebSocketTarget` error which must be again handled by the `errorCaught` callback of the `HTTPHandler`. 

Note that we are handling this error within the HTTP server. We don't upgrade if this error is thrown.


pushkarnk:
@nethraravindran Can you review this please? Thanks.

nethraravindran:
@pushkarnk  This looks good to me :)

Issue #67, opened 2018-08-16
Kitura tests on kitura-nio branch leak the MultiThreadedEventLoopGroup threads


weissi:
This might really be a kitura issue but it probably also happens here: Kitura tests on kitura-nio branch leak the MultiThreadedEventLoopGroup threads, basically the `MultiThreadedEventLoopGroup` that are created in the tests are never torn down, ie. the threads are still kept alive.

Essentially it's the [same issue that Vapor has too](https://github.com/vapor/vapor/issues/1778)

[`MultiThreadedEventLoopGroup.syncShutdownGracefully`](https://apple.github.io/swift-nio/docs/current/NIO/Classes/MultiThreadedEventLoopGroup.html#/s:3NIO14EventLoopGroupP18shutdownGracefullyy8Dispatch0G5QueueC5queue_ys5Error_pSgctF) should be used for example in the `tearDown` method of a test

nethraravindran:
@pushkarnk I guess we are already doing that https://github.com/IBM-Swift/Kitura-NIO/blob/106677541e0bcd13c3a631db15427e1e8c614235/Sources/KituraNet/HTTP/HTTPServer.swift#L222 here.

weissi:
@pushkarnk/@nethraravindran I think it might be the one from `ClientRequest`.

nethraravindran:
@weissi Thank you for your input. Yes it is from `ClientRequest`.  Now, with `syncShutdownGracefully` in ClientRequest I see a very minimal leak :)

nethraravindran:
PR https://github.com/IBM-Swift/Kitura-NIO/pull/76

@weissi  Thanks! With this fix, I see no leak :)

weissi:
awesome, thanks 👍. I'm going to close this

PR #76, opened 2018-08-21
Shut down the eventloop gracefully in ClientRequest

nethraravindran:
ClientRequest leaks threads because we don't shutdown the MultiThreadedEventLoop group associated with it. This PR resolves the leak by shutting down every `MultiThreadedEventLoopGroup` that were created in the application using `syncShutdownGracefully()`

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -380,6 +380,14 @@ public class ClientRequest {
             self.port = isHTTPS ? 443 : 80
         }
 
+        defer {
+            do {
+                try group.syncShutdownGracefully()
+            } catch {
+                Log.error("ClientRequest failed to shut down the EventLoopGroup for the requested URL: \(url)")
+            }
+        }
+
         do {
             channel = try bootstrap.connect(host: hostName, port: Int(self.port ?? 80)).wait()
         } catch let error {


nethraravindran:
Fixes https://github.com/IBM-Swift/Kitura-NIO/issues/67

Sources/KituraNet/ClientRequest.swift
```diff
@@ -380,6 +380,10 @@ public class ClientRequest {
             self.port = isHTTPS ? 443 : 80
         }
 
+        defer {
+            try! group.syncShutdownGracefully()
```
ianpartridge: Are we sure we want to crash the app if the threads don't shut down (for whatever reason)? Is there an argument for carrying on?

```swift
do {
    try group.syncShutdownGracefully()
} catch {
    // meh
}

Comment (pushkarnk) [reply Sources/KituraNet/ClientRequest.swift]: I think we should just log a diagnostic message and carry on here. We don't want to crash.

**Comment by nethraravindran** [comment]
@pushkarnk I have addressed the review comments. Please review.

File: Sources/KituraNet/ClientRequest.swift [hunk 2]
```diff
@@ -380,6 +380,14 @@ public class ClientRequest {
             self.port = isHTTPS ? 443 : 80
         }
 
+        defer {
+            do {
+                try group.syncShutdownGracefully()
+            } catch {
+                Log.error("Failed to shut down the EventLoopGroup")
```
ianpartridge: Please include "ClientRequest" and maybe the URL for the request?

nethraravindran: @ianpartridge Sure! Thanks :)

Issue #62, opened 2018-08-13
Linker error in libz on Ubuntu 16.04


nethraravindran:
```
Linking ./.build/x86_64-unknown-linux/debug/KituraPackageTests.xctest
/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:function _T08NIOHTTP122HTTPResponseCompressorC19deinitializeEncoder33_E8A369BA8CC72FD6D05117C456504CA2LLyyF: error: undefined reference to 'deflateEnd'
/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:function _T08NIOHTTP119PartialHTTPResponse33_E8A369BA8CC72FD6D05117C456504CA2LLV12compressBody3NIO10ByteBufferVSgSC10z_stream_sVz10compressor_AF0oP9AllocatorV9allocators5Int32V4flagtF: error: undefined reference to 'deflateBound'
/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:function _T0SC10z_stream_sV8NIOHTTP1E15deflateToBuffer33_E8A369BA8CC72FD6D05117C456504CA2LLs5Int32V3NIO04ByteE0Vz6buffer_AG4flagtFSis016UnsafeMutableRawE7PointerVcfU_: error: undefined reference to 'deflate'
/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:/root/Kitura/.build/x86_64-unknown-linux/debug/NIOHTTP1.build/HTTPResponseCompressor.swift.o:function CNIOZlib_deflateInit2: error: undefined reference to 'deflateInit2_'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
<unknown>:0: error: link command failed with exit code 1 (use -v to see invocation)
```

[...]

PR #77, opened 2018-08-21
Install pkg-config on Ubuntu 16.04 docker

nethraravindran:
`pkg-config` is not installed on Ubuntu 16.04 docker. And hence we get linker error in libz. Installing `pkg-config` resolves the issue.


nethraravindran:
Fixes [https://github.com/IBM-Swift/Kitura-NIO/issues/62](url)

PR #78, opened 2018-08-23
Run the Kitura router on a DispatchQueue instead of an ELT

pushkarnk:
Currently, the Kitura router that eventually invokes the endpoint handlers runs on the event loop threads. This is bad design primarily because the handlers may invoke blocking code which will end up blocking the event loop threads, reducing their availability to handle newer HTTP requests.

This pull request proposes moving the Kitura router invocation to a DispatchQueue, effectively freeing up the event loop thread which was scheduled to handle the related HTTP request. We move back to the event loop while writing the response - `HTTPServerResponse.end()`. Error handling (error responses) is done on the event loop. To write responses we use `Channel`s,  instead of `ChannelHandlerContext`s, owing to the thread-safety provided by them.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPHandler.swift
+++ b/Sources/KituraNet/HTTP/HTTPHandler.swift
@@ -19,6 +19,7 @@ import NIOHTTP1
 import NIOWebSocket
 import LoggerAPI
 import Foundation
+import Dispatch
 
 public class HTTPHandler: ChannelInboundHandler {
 
@@ -72,12 +73,13 @@ public class HTTPHandler: ChannelInboundHandler {
                 serverRequest.buffer!.byteBuffer.write(buffer: &buffer)
             }
         case .end(_):
-            serverResponse = HTTPServerResponse(ctx: ctx, handler: self)
+            serverResponse = HTTPServerResponse(channel: ctx.channel, handler: self)
             //Make sure we use the latest delegate registered with the server
-            if let delegate = server.delegate {
-                Monitor.delegate?.started(request: serverRequest, response: serverResponse)
-                delegate.handle(request: serverRequest, response: serverResponse)
-            } //TODO: failure path
+            DispatchQueue.global().async {
+                let delegate = self.server.delegate ?? HTTPDummyServerDelegate()
+                Monitor.delegate?.started(request: self.serverRequest, response: self.serverResponse)
+                delegate.handle(request: self.serverRequest, response: self.serverResponse)
+            }
         }
     }
 
@@ -110,16 +112,13 @@ public class HTTPHandler: ChannelInboundHandler {
             let target = server.latestWebSocketURI ?? "/<unknown>"
             message = "No service has been registered for the path \(target)"
         default:
-            if error is HTTPParserError {
-                break
-            }
-            // Don't handle any other errors for now!
+            // Don't handle any other errors, including `HTTPParserError`s
             return
         }
 
         do {
+            serverResponse = HTTPServerResponse(channel: ctx.channel, handler: self)
             errorResponseSent = true
-            serverResponse = HTTPServerResponse(ctx: ctx, handler: self)
             try serverResponse.end(with: .badRequest, message: message)
         } catch {
             Log.error("Failed to send error response")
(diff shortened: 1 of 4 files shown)


Sources/KituraNet/HTTP/HTTPServerResponse.swift
```diff
@@ -23,7 +23,7 @@ import Foundation
 public class HTTPServerResponse: ServerResponse {
 
     /// The channel handler context on which an HTTP response should be written
-    private weak var ctx: ChannelHandlerContext?
+    private weak var channel: Channel?
```
ianpartridge: The comment above needs editing.

Sources/KituraNet/HTTP/HTTPHandler.swift
```diff
@@ -72,12 +73,15 @@ public class HTTPHandler: ChannelInboundHandler {
...
-            if let delegate = server.delegate {
-                Monitor.delegate?.started(request: serverRequest, response: serverResponse)
-                delegate.handle(request: serverRequest, response: serverResponse)
-            } //TODO: failure path
+            DispatchQueue.global().async { [weak self] in
+                guard let weakSelf = self else { fatalError("No HTTPHandler to proceed with the response.") }
+                if let delegate = weakSelf.server.delegate {
+                    Monitor.delegate?.started(request: weakSelf.serverRequest, response: weakSelf.serverResponse)
+                    delegate.handle(request: weakSelf.serverRequest, response: weakSelf.serverResponse)
+                } //TODO: failure path
```
ianpartridge: Can we sort out this failure TODO now?

Sources/KituraNet/HTTP/HTTPHandler.swift
```diff
@@ -72,12 +73,15 @@ public class HTTPHandler: ChannelInboundHandler {
...
         case .end(_):
-            serverResponse = HTTPServerResponse(ctx: ctx, handler: self)
+            serverResponse = HTTPServerResponse(channel: ctx.channel, handler: self)
             //Make sure we use the latest delegate registered with the server
-            if let delegate = server.delegate {
-                Monitor.delegate?.started(request: serverRequest, response: serverResponse)
-                delegate.handle(request: serverRequest, response: serverResponse)
-            } //TODO: failure path
+            DispatchQueue.global().async { [weak self] in
+                guard let weakSelf = self else { fatalError("No HTTPHandler to proceed with the response.") }
```
ianpartridge: Is fatalError() essential or could we carry on?

Tests/KituraNetTests/RegressionTests.swift
```diff
@@ -166,28 +166,26 @@ class RegressionTests: KituraNetTest {
...
-            var goodClient = GoodClient(with: HTTPClient(with: self.expectation(description: "Bad request error")))
-            try! goodClient.makeGoodRequestFollowedByBadRequest(serverPort)
-            waitForExpectations(timeout: 10)
-
-        } catch {
-            XCTFail("Couldn't start server")
-        }
-
-    }
+    //func testBadRequestFollowingGoodRequest() {
```
ianpartridge: We don't need to comment out the code of the test, just the entry in the test array.

pushkarnk: I think that doesn't work on macOS. I think Xcode doesn't pick the tests from the array, it runs all tests found in a file. I may be wrong - just trying a travis CI run.

Sources/KituraNet/HTTP/HTTPHandler.swift
pushkarnk: Right, we don't need a `fatalError` there. I have replaced it with a `Log.error`. I think there's a lot of other places that need this change. Will do a cleanup PR soon.

Sources/KituraNet/HTTP/HTTPHandler.swift
pushkarnk: Yeah, I think we can use the `HTTPDummyServerDelegate` here. We use that if no delegate is registered.

pushkarnk:
@ianpartridge Thanks for the review! I've addressed all of your comments.

Tests/KituraNetTests/RegressionTests.swift
ianpartridge: I'm sorry, you were right!

Issue #2, opened 2018-05-14
Bind errors while running Kitura and Kitura-NIO tests


pushkarnk:
With SwiftNIO, Channel.close() calls close(socket) which moves the underlying socket to the TIMED_WAIT state. In this state, subsequent bind calls fail, until the socket moves out of the TIMED_WAIT state and is actually closed. Kitura tests fail frequently with this error. While using Kitura-net that uses BlueSocket, I can see a shutdown(socket) being issued before close(socket). That may be the reason why this failure isn't seen with KituraNet. Needs investigation.

We can workaround this using `HTTPServer.allowPortReuse = true`

pushkarnk:
We should try using a ephemeral ports for all the tests.

weissi:
@pushkarnk setting 

      .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)

should fix that. I reckon KituraNet just sets that automatically whilst NIO wants the user to specify it explicity.

ianpartridge:
@pushkarnk ^ did you see this?

pushkarnk:
@ianpartridge Yes, we tried `SO_REUSEADDR` in the past and it doesn't solve this problem. 

@weissi I remember discussing this issue with you over Slack back in May. `SO_REUSEADDR` wasn't helping us and you suggested we try `SO_REUSEPORT`.

pushkarnk:
@weissi Now I recall it better, we needed `SO_REUSEPORT` for a particular test called `testServersSharingPort`. Looks like `SO_REUSEADDR` is enough for this issue.

weissi:
@pushkarnk yes, `testServersSharingPort` sounds like two servers bind on the same port which would be `SO_REUSEPORT` (and a new enough Linux version)

weissi:
in any case, I'd recommend `SO_REUSEADDR` for all tests and `SO_REUSEPORT` for the ones where you actually need to have two (or more) servers bind to _the same port at the same time_.

pushkarnk:
@weiss This is the gist of the problem we're facing using both `SO_REUSEADDR` and `SO_REUSEPORT`

Consider this NIO program (note the commented line of code):
```
import NIO 

let bootstrap1 = ServerBootstrap(group: MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount))
[...]

PR #82, opened 2018-08-29
Set SO_REUSEADDR for all socket channels to avoid bind errors

nethraravindran:
**Motivation:**
To allow reuse socket address and port.

Previously swift-nio was overriding `ChannelOptions.socket` ie. allowing only one socket option to set. 

Hence as a work-around we were always reusing the port based on the value of `allowPortReuse` 

With this fix https://github.com/apple/swift-nio/pull/597 , swift-nio allows to set two distinct `ChannelOptions.socket` of one type.  

And so we can have both the options `ChannelOptions.socket` set for the server
```
.serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT), value: 1)
.serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
```

With this we reuse the port only when necessary.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -130,6 +130,7 @@ public class HTTPServer : Server {
 
         let bootstrap = ServerBootstrap(group: eventLoopGroup)
             .serverChannelOption(ChannelOptions.backlog, value: BacklogOption.OptionType(self.maxPendingConnections))
+            .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -28,7 +28,7 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(url: "https://github.com/apple/swift-nio.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", .branch("master")),
```
nethraravindran: This is a temporary change until we have swift-nio 1.9.3

ianpartridge: [1.9.3](https://github.com/apple/swift-nio/releases/tag/1.9.3) is out :)

nethraravindran: @ianpartridge Thanks! I'm changing it now

nethraravindran:
Fixes [https://github.com/IBM-Swift/Kitura-NIO/issues/2](url)

PR #83, opened 2018-08-30
Execute tasks on the event loop - refactor

nethraravindran:


Sources/KituraNet/ClientRequest.swift
```diff
@@ -404,23 +404,26 @@ public class ClientRequest {
...
-            } else {
-                channel.eventLoop.execute {
-                    self.channel.close(promise: nil)
-                }
+            }
+        }
+    }
+
+    /// Executes task on event loop
+    private func runOnEventLoop(channel: Channel, _ task: @escaping () -> Void) {
```
ianpartridge: Can we do the same `execute(on: channel.eventLoop)` here and reuse code?

Sources/KituraNet/HTTP/HTTPServerResponse.swift
```diff
@@ -115,13 +118,14 @@ public class HTTPServerResponse: ServerResponse {
...
-            try end0(channel: channel)
-        } else {
-            channel.eventLoop.execute {
-                try! self.end0(channel: channel)
+
+        execute(on: channel.eventLoop) {
+            do {
+                try self.end0(channel: channel)
+            } catch let error {
+                fatalError("Error: \(error)")
```
nethraravindran: @pushkarnk  should we crash here?

pushkarnk: I don't think we should crash at these places. We must simply `Log.error()` and return. We need to do all of that in a separate PR though.

PR #84, opened 2018-08-30
Improvements in HTTPServerResponse.write() methods

pushkarnk:
This commit includes a fix for an embarrasing bug - we wrote to the response buffer only if it was nil, after allocating it. This meant, for N consecutive writes to it, only the first would be honored and N-1 would be ignored. This change also eagerly allocates the response buffer with an initial size of 2000, just like KituraNet does it. We don't keep the buffer as an optional anymore. We allocate it in the initializer because we are pretty sure that the HTTPServerResponse initializer always runs on an event loop.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -52,11 +52,15 @@ public class HTTPServerResponse: ServerResponse {
     private var httpVersion: HTTPVersion
 
     /// The data to be written as a part of the response.
-    private var buffer: ByteBuffer?
+    private var buffer: ByteBuffer
+
+    /// Initial size of the response buffer (inherited from KituraNet)
+    private static let bufferSize = 2000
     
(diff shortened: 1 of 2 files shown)


ianpartridge:
Test please :-)

pushkarnk:
Thanks for the review @ianpartridge ! I've added a test

pushkarnk:
Found one more possibility in the code without the changes in this PR. Consider a response that writes twice to the ByteByffer. Because we submit the write task (creating a buffer and writing to it) on the event loop, the first write task may not have been invoked when the second one is about to be submitted. This results in a double allocation of the ByteBuffer, the first buffer being discarded. This pull request fixes that scenario too. 

P.S: A glance at the implementation of `NIO.SelectableEventLoop.execute()` shows that we do impose an order among  the submitted tasks, its a FIFO.

ianpartridge:
Tag a patch release?

pushkarnk:
@ianpartridge tagged v1.0.4

PR #86, opened 2018-09-06
Refactor - HTTPServerResponse.end() methods

nethraravindran:
Refactored and cleaned up the code around `end()` methods

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -41,12 +41,11 @@ public class HTTPServerResponse: ServerResponse {
             if let newValue = newValue {
                 status = newValue.rawValue
             }
-        } 
+        }
     }
 
     /// The HTTP headers to be sent to the client as part of the response.
     public var headers : HeadersContainer = HeadersContainer()
-
 
     /// The HTTP version to be sent in the response.
     private var httpVersion: HTTPVersion
@@ -56,7 +55,7 @@ public class HTTPServerResponse: ServerResponse {
 
     /// Initial size of the response buffer (inherited from KituraNet)
     private static let bufferSize = 2000
-    
+
     init(channel: Channel, handler: HTTPRequestHandler) {
         self.channel = channel
         self.handler = handler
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTPServerResponse.swift
```diff
@@ -180,11 +156,26 @@ public class HTTPServerResponse: ServerResponse {
...
+        }
+    }
+
+    /// Send response to the client
+    private func sendResponse(channel: Channel, handler: HTTPRequestHandler, status: HTTPResponseStatus, withBody: Bool = false) throws {
+        let response = HTTPResponseHead(version: httpVersion, status: status, headers: headers.httpHeaders())
         channel.write(handler.wrapOutboundOut(.head(response)), promise: nil)
         if withBody && buffer.readableBytes > 0 {
             channel.write(handler.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: nil)
+        } else if buffer.readableBytes > 0 {
```
pushkarnk: This case may not be necessary. We must be able to cover everything in the first `if` case.

Sources/KituraNet/HTTP/HTTPServerResponse.swift
```diff
@@ -180,11 +156,26 @@ public class HTTPServerResponse: ServerResponse {
...
+            do {
+                try self.sendResponse(channel: channel, handler: handler, status: status, withBody: withBody)
+            } catch let error {
+                fatalError("Error: \(error)")
+            }
+        }
+    }
+
+    /// Send response to the client
+    private func sendResponse(channel: Channel, handler: HTTPRequestHandler, status: HTTPResponseStatus, withBody: Bool = false) throws {
```
pushkarnk: `sendResponse` is going to be used for both normal and error responses. So, it makes sense to have `withBody` defaulting to `true`. We can assume that most of the responses do have a body.

nethraravindran: @pushkarnk Thanks! I have addressed the changes

pushkarnk:
Thanks!

PR #92, opened 2018-09-17
Refactor WebSocketUpgrader configuration to improve readability

nethraravindran:
Refactoring this enhances the readability of upgrade handlers in HTTPServer

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -114,6 +114,21 @@ public class HTTPServer : Server {
         return ctx.channel.pipeline.add(handler: webSocketHandlerFactory.handler(for: serverRequest))
     }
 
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -114,6 +114,21 @@ public class HTTPServer : Server {
         return ctx.channel.pipeline.add(handler: webSocketHandlerFactory.handler(for: serverRequest))
     }
 
+    typealias ShouldUpgradeFunction = (HTTPRequestHead) -> HTTPHeaders?
+    typealias UpgradePipelineHandlerFunction = (Channel, HTTPRequestHead) -> EventLoopFuture<Void>
```
pushkarnk: Does it make sense to have these as `private` typealiases for now?

nethraravindran: Yes makes sense. Thanks :)

PR #95, opened 2018-09-18
Remove forced uwrapped optional try(s)

nethraravindran:
The motive is to remove forced unwrapped optional `try` methods and replace with `do-try-catch ` to avoid crash

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -147,7 +147,11 @@ public class HTTPServer : Server {
         self.port = port
 
         if let tlsConfig = tlsConfig {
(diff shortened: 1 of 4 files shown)


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -245,14 +249,22 @@ public class HTTPServer : Server {
...
     }
 
     /// Stop listening for new connections.
     public func stop() {
         guard serverChannel != nil else { return }
-        try! serverChannel.close().wait()
+        do {
+            try serverChannel.close().wait()
+        } catch let error {
+            Log.error("Error: \(error)")
```
pushkarnk: Can we have a more detailed error, something like `Failed to close the server channel. Error: \(error)`

PR #96, opened 2018-09-24
Do not store the HTTPDummyServerDelegate object

pushkarnk:
Kitura-Net keeps the HTTP server running even if a delegate isn't provided. We use a HTTPDummyServerDelegate for this purpose. Before this code change, we stored the dummy delegate in the delegate property of HTTPServer. This may be confusing to a user who will rightly think that delegate is nil if it isn't explicitly set.


pushkarnk:
Ran Kitura tests locally

nethraravindran:
Looks good to me

PR #97, opened 2018-09-25
Remove installation for libressl from travis.yml

nethraravindran:
`swift-nio-ssl` requires installing libressl on OSX. This has been directly added in the Package builder https://github.com/IBM-Swift/Package-Builder/pull/141. Hence we can remove it from Kitura-NIO


pushkarnk:
With[ this change](url
https://github.com/IBM-Swift/Package-Builder/pull/141) to Package-Builder, the PR build is going through. I think this is good to merge now.

PR #103, opened 2018-10-24
Use ephemeral ports in tests

pushkarnk:
The tests that create an HTTP server always bind to and listen on port number 8080. Though this is fine when we run the tests sequentially, we run into port collisions when we try to run them in parallel (using `swift test --parallel`). We must ideally be using ephemeral ports to avoid this issue.

The other option is to configure SO_REUSEPORT by default, but with this tests that verify the presence  and absence of SO_REUSEPORT become irrelevant. 

Ephemeral ports may lie in the range starting 16000 and up to 65535. Using the `sysctl` command on Linux and macOS, it is possible to limit them to a given range. In the context of Kitura, limiting them to a range so that the port number can be accommodated in an `UInt16` essential because `ClientRequest.Options.port` is an `UInt16`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,17 +26,17 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=ubuntu:16.04
+      env: DOCKER_IMAGE=ubuntu:16.04 DOCKER_PRIVILEGED=true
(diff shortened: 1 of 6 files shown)


pushkarnk:
All the tests seem to be passing now. I'm merging this. Fingers crossed.

PR #106, opened 2018-11-13
Reuse HTTPServerRequest and HTTPServerResponse objects

pushkarnk:
This pull request deals with an optimisation in the context of a special case. It is likely that identical HTTP requests are sent in succession over an HTTP connection. In that case, Kitura-NIO must be able to reuse the request and response abstraction objects (of course, suitably reset).

Reusing HTTPServerRequest objects can make it possible to reuse some of the computed data, by simply caching it, instead of repeating the same computation. See [this commit](https://github.com/IBM-Swift/Kitura-NIO/commit/664709013da1729302ce1787649b05a1202eaf2b) and [this commit](https://github.com/IBM-Swift/Kitura-NIO/commit/147623f19372c579c56c83c5f0566d4fe1937749), for example.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPRequestHandler.swift
+++ b/Sources/KituraNet/HTTP/HTTPRequestHandler.swift
@@ -70,6 +70,8 @@ internal class HTTPRequestHandler: ChannelInboundHandler {
     public typealias InboundIn = HTTPServerRequestPart
     public typealias OutboundOut = HTTPServerResponsePart
 
+    private var prevHTTPRequestHead: HTTPRequestHead? = nil
+
     public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
         let request = self.unwrapInboundIn(data)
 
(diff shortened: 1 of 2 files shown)


pushkarnk:
An obvious question here is:  Why don't we maintain a bounded dictionary of HTTPRequestHead as keys and HTTPServerRequest as values, over a connection? This tends to have a better scope than the approach in this pull request, which just takes care of identical requests that are subsequent.

The answer to that is: Though one can have HTTPRequestHead conform to Hashable, maintaining a dictionary seems to have a large overhead. The performance dips by 7-9%.

pushkarnk:
I just now realised that it is easier to reuse the request and response objects if we are running the handlers on the event loops. Because Kitura-NIO runs them asynchronously on a parallel Dispatch queue, we'd need to make sure the reuse doesn't cause race conditions. I'm working on this.

djones6:
Rather than a dictionary, could we maintain a (short) list of RequestHeads? This would not require Hashable, instead we'd compare to (say) the last 5 distinct heads seen.

The size of that list would be a trade-off between performance overhead of comparing multiple heads, vs the gain and likelihood of finding a match.

pushkarnk:
Thanks for the feedback @djones6 

[Here](https://github.com/pushkarnk/Kitura-NIO/commit/57801cae9c3af5f0037f95b7cb707242ba58ca25) is a change based on your feedback. The problem with using a bounded array is the overhead of its management. Do you think this is acceptable?

PR #113, opened 2018-11-22
Do headers translation only once

pushkarnk:
Translating HTTPHeaders to HeadersContainer is an inevitable and expensive task. Given that the HTTPHeaders from HTTPRequestHead will not change, the translation could be a one-time thing. Currently, we do the translation everytime the `headers` property is accessed. We could do an eager translation in the initializer or a lazy translation when the property is accessed for the first time. Given that Kitura will most likely access the headers at least once, an eager translation shouldn't be a problem here.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -24,11 +24,7 @@ import Foundation
 public class HTTPServerRequest: ServerRequest {
 
     /// Set of HTTP headers of the request.
-    public var headers : HeadersContainer {
-        return HeadersContainer.create(from: _headers)
-    }
-
-    private var _headers: HTTPHeaders
+    public var headers : HeadersContainer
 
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -24,9 +24,7 @@ import Foundation
 public class HTTPServerRequest: ServerRequest {
 
     /// Set of HTTP headers of the request.
-    public var headers : HeadersContainer {
-        return HeadersContainer.create(from: _headers)
-    }
+    public var headers : HeadersContainer
 
     private var _headers: HTTPHeaders
```
nethraravindran: Should we have `_headers` as well?

nethraravindran: The only place where we use is to get `Host`. Can we now re-use `headers.httpHeader()` to obtain the same if it is not expensive?

pushkarnk: @nethraravindran Yes, I agree. I've made the change.

PR #120, opened 2018-12-17
Initialize ELG with the number of affinitized cores

pushkarnk:
The MultiThreadedEventLoopGroup that runs the HTTPServer is initialized with all the available cores. On Linux, it could be instead initialised to the number of affinitized cores. If the number of affinitized cores is lesser than the `System.coreCount`, initializing the MultiThreadedEventLoopGroup with the later comes with a performance penalty.


pushkarnk:
@djones6 Kindly review. The only premise for this pull request is the performance penalty that comes along if the `number of _affinitised_ cores` is lesser than the `total core count` and we initialise the MultiThreadedEventLoopGroup with the latter. Thanks!

PR #123, opened 2018-12-24
Remove forced unwrapped optional try(s)

nethraravindran:


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -208,7 +208,11 @@ public class HTTPServer: Server {
 
         let queuedBlock = DispatchWorkItem(block: {
             guard let serverChannel = self.serverChannel else { return }
-            try! serverChannel.closeFuture.wait()
+            do {
+                try serverChannel.closeFuture.wait()
+            } catch let error {
+                Log.error("Server channel has been closed. Error: \(error)")
```
pushkarnk: In my understanding, if `closeFuture.wait()` throws an error, it must be logged as a `close error`. So, this message could be something like `"Error while closing channel: \(error)"`

pushkarnk: @nethraravindran

pushkarnk:
Thanks. Merging this now.

PR #124, opened 2018-12-24
Fix: SwiftLint warnings

nethraravindran:


nethraravindran:
@pushkarnk Should we consider modifying/increasing the limit?

Issue #101, opened 2018-10-19
flaky test ClientE2ETests.testSimpleHTTPClient


weissi:
## failure

```
Test Case 'ClientE2ETests.testSimpleHTTPClient' started at 2018-10-19 15:44:10.365
/tmp/.nio-release-tools_g7eLXu/Kitura-NIO/Tests/KituraNetTests/ClientE2ETests.swift:113: error: ClientE2ETests.testSimpleHTTPClient : XCTAssertEqual failed: ("Optional(KituraNet.HTTPStatusCode.serviceUnavailable)") is not equal to ("Optional(KituraNet.HTTPStatusCode.OK)") - HTTP Status code was Optional(KituraNet.HTTPStatusCode.serviceUnavailable)
Test Case 'ClientE2ETests.testSimpleHTTPClient' failed (30.195 seconds)
```

## OS

Ubuntu 14.04 in Docker

## code

KituraNIO hash: 214abc91b6fd51b7d2817175f0599ad120754ce6
swift-nio: 9bae007a15b49d7fe685b6b33308822c6d0536c0
swift-nio-ssl: db16c3a90b101bb53b26a58867a344ad428072e0

```
$ cat Package.resolved 
{
  "object": {
    "pins": [
      {
        "package": "Socket",
        "repositoryURL": "https://github.com/IBM-Swift/BlueSocket.git",
        "state": {
          "branch": null,
          "revision": "fb4e74cd8eb3ce4a2b860fbf727b2dada796d77d",
          "version": "1.0.22"
        }
      },
      {
        "package": "SSLService",
        "repositoryURL": "https://github.com/IBM-Swift/BlueSSLService.git",
        "state": {
          "branch": null,
          "revision": "e2af36601f91b3cefa371c00ecdcb62fef1d000a",
          "version": "1.0.22"
        }
      },
      {
        "package": "LoggerAPI",
        "repositoryURL": "https://github.com/IBM-Swift/LoggerAPI.git",
        "state": {
          "branch": null,
          "revision": "5041f2673aa75d6e973d9b6bd3956bc5068387c8",
          "version": "1.7.3"
        }
      },
      {
        "package": "OpenSSL",
        "repositoryURL": "https://github.com/IBM-Swift/OpenSSL.git",
        "state": {
          "branch": null,
          "revision": "ba74a683141fde66d4bf88358b6a678d7e5a27ce",
          "version": "1.0.1"
        }
      },
      {
[...]

PR #131, opened 2018-12-29
Fix for flaky test ClientE2ETests.testSimpleHTTPClient

nethraravindran:
Fixes https://github.com/IBM-Swift/Kitura-NIO/issues/101

### DIFF
--- a/Tests/KituraNetTests/ClientE2ETests.swift
+++ b/Tests/KituraNetTests/ClientE2ETests.swift
@@ -103,23 +103,25 @@ class ClientE2ETests: KituraNetTest {
     }
 
     func testSimpleHTTPClient() {
-        let expectation = self.expectation(description: "HTTP client response completed")
-        let url = "http://httpbin.org"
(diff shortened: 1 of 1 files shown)


Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -335,4 +337,19 @@ class ClientE2ETests: KituraNetTest {
...
+
+        func handle(request: ServerRequest, response: ServerResponse) {
+            do {
+                let result: String = "Hello, World!"
+                response.statusCode = .OK
+                response.headers["Content-Type"] = ["text/plain"]
+                response.headers["Content-Length"] = ["\(result.count)"]
+                try response.end(text: result)
+            } catch {
+                print("Error while writing response")
```
pushkarnk: Can we have an `XCTFail(...)` here instead of the `print(...)`? Its better to fail early.

nethraravindran: Sure @pushkarnk

PR #135, opened 2019-01-03
HTTPServerRequest: initialize localAddress and remoteAddress lazily

pushkarnk:
In `HTTPServerRequest` private vars `localAddress` and `localAddressPort` are used only in the implementation of the `urlURL` property. However, they are always initialised in the `HTTPServerRequest` initialiser. We can move their initialisation within the `urlURL` implementation.

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -398,26 +398,15 @@ public class ClientRequest {
         // Make the HTTP request, the response callbacks will be received on the HTTPClientHandler.
         // We are mostly not running on the event loop. Let's make sure we send the request over the event loop.
         guard let channel = channel else { return }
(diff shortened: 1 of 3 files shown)


pushkarnk:
`ChanneHandlerContext.localAddress` may be accessed only one the event loop associated with the channel handler context. This change isn't feasible. Closing.

pushkarnk:
@nethraravindran I've pushed a commit that initialises `remoteAddress` in a lazy manner

pushkarnk:
@nethraravindran I've pushed a few more changes. Could you please review the new patch and approve? Thanks!

nethraravindran:
@pushkarnk I guess we can now re-use `func run(_ task: @escaping () -> Void) -> EventLoopFuture<Void>?` in `ClientRequest` as well

PR #139, opened 2019-01-04
Support for Code Climate

nethraravindran:


pushkarnk:
We aren't really using the output of codebeat. Closing this PR after discussing the same with @nethraravindran

PR #140, opened 2019-01-04
tests: Use ephemeral ports by default

pushkarnk:
Instead of restricting the ephemeral port range to accommodate port numbers in an Int16 (a known limitation of `ClientRequest`), we can cast Int16s to UInt16s and vice versa. This removes the need to have the USE_EPHEMERAL_PORTS compiler flag and the script that issues commands to restrict the ephemeral port range.

### DIFF
--- a/.run-tests-parallel
+++ b/.run-tests-parallel
@@ -1,1 +1,1 @@
-swift test -Xswiftc -DUSE_EPHEMERAL_PORTS --parallel
+swift test --parallel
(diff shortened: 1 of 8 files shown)


djones6:
Nice solution!

pushkarnk:
Thanks to @weissi for suggesting this!

pushkarnk:
Thanks for the review @djones6 I've pushed a commit with the change you suggested.

djones6:
@pushkarnk This looks good to me.  Once this is merged and tagged, you can update https://github.com/IBM-Swift/Kitura/pull/1382 accordingly.

(The Kitura-net equivalent, https://github.com/IBM-Swift/Kitura-net/pull/285, has been merged and tagged).

PR #143, opened 2019-01-08
HTTPServerRequest: Build urlURL using URLComponents

nethraravindran:
This increases the performance by ~2% while accessing`urlURL`

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -41,55 +41,61 @@ public class HTTPServerRequest: ServerRequest {
 
     @available(*, deprecated, message: "URLComponents has a memory leak on linux as of swift 3.0.1. use 'urlURL' instead")
     public var urlComponents: URLComponents {
-        return URLComponents(url: urlURL, resolvingAgainstBaseURL: false) ?? URLComponents()
+        if let _urlComponents = _urlComponents {
+            return _urlComponents
+        }
+        _urlComponents = URLComponents(url: urlURL, resolvingAgainstBaseURL: false) ?? URLComponents()
+        return _urlComponents!
     }
 
     private var _url: URL?
+
+    private var _urlComponents: URLComponents?
 
     public var urlURL: URL {
         if let _url = _url {
             return _url
         }
-        var url = ""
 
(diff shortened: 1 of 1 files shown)


pushkarnk:
@nethraravindran Bridging UInt16 to NSNumber doesn't work pre-4.2. We have two options:
1. replace it by `NSNumber(value:)` all throughout
2. do (1) only for pre-4.2 compiler versions

I ran a small benchmark and find performance of `NSNumber(value:)` similar to `x as NSNumber`. So, we may adopt (1).

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -41,18 +41,24 @@ public class HTTPServerRequest: ServerRequest {
 
     @available(*, deprecated, message: "URLComponents has a memory leak on linux as of swift 3.0.1. use 'urlURL' instead")
     public var urlComponents: URLComponents {
-        return URLComponents(url: urlURL, resolvingAgainstBaseURL: false) ?? URLComponents()
+        if let _urlComponents = _urlComponents {
+            return _urlComponents
+        } else {
+            return URLComponents(url: urlURL, resolvingAgainstBaseURL: false) ?? URLComponents()
```
pushkarnk: @nethraravindran We must cache it here as well, in the scenario where `urlComponents` was invoked before `urlURL`

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -41,18 +41,24 @@ public class HTTPServerRequest: ServerRequest {
...
+    private var _urlComponents: URLComponents?
+
     public var urlURL: URL {
         if let _url = _url {
             return _url
         }
-        var url = ""
 
-        self.enableSSL ? url.append("https://") : url.append("http://")
+        _urlComponents = URLComponents()
```
pushkarnk: check if `_urlComponents` already exists here, we could have initialised it in `urlComponents`

Sources/KituraNet/HTTP/HTTPServerRequest.swift
nethraravindran: @pushkarnk Sure!

PR #146, opened 2019-01-08
Lazily initialise the properties of URLParser

nethraravindran:
This increases the performance  by 5-7% (approx)

### DIFF
--- a/Sources/KituraNet/URLParser.swift
+++ b/Sources/KituraNet/URLParser.swift
@@ -22,29 +22,92 @@ import Foundation
 
 public class URLParser: CustomStringConvertible {
 
+    private var urlComponents: URLComponents?
+
     /// Schema.
-    public var schema: String?
+    public var schema: String? {
+        get {
+            return self.urlComponents?.scheme
+        }
+        set (newValue) {
+            self.urlComponents?.scheme = newValue
+        }
+    }
 
     /// Hostname.
-    public var host: String?
+    public var host: String? {
+        get {
+            return self.urlComponents?.host
+        }
+        set (newValue) {
+            self.urlComponents?.host = newValue
+        }
+    }
 
     /// Path portion of the URL.
-    public var path: String?
+    public var path: String? {
+        get {
+            return self.urlComponents?.path
+        }
+        set (newValue) {
+            if let newValue = newValue {
+                self.urlComponents?.path = newValue
+            }
+        }
+    }
 
     /// The entire query portion of the URL.
-    public var query: String?
+    public var query: String? {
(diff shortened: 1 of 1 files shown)


pushkarnk:
@nethraravindran I think all of the properties of `URLParser` are read-write. Can you write setters for each of them here (except `description`) ?

nethraravindran:
@pushkarnk Thank you! I have addressed the review comments.

Sources/KituraNet/URLParser.swift
```diff
@@ -22,29 +22,113 @@ import Foundation
...
+            _userInfo = "\(username):\(password)"
+        }
+        return _userInfo
+    }()
 
     /// The userid and password if specified in the URL.
-    public var userinfo: String?
+    public var userinfo: String? {
+        get {
+            print("inside get")
```
ianpartridge: Remove this.

nethraravindran:
@ianpartridge Thanks for the review!  I've addressed the changes.  But`userinfo` and `queryParameters` can be get only properties since they depend on other parameters

Sources/KituraNet/URLParser.swift
```diff
@@ -22,29 +22,89 @@ import Foundation
...
+        }
+        set (newValue) {
+            self.urlComponents?.fragment = newValue
+        }
+    }
 
     /// The userid and password if specified in the URL.
-    public var userinfo: String?
+    public var userinfo: String? {
+        var _userInfo: String? = nil
```
ianpartridge: You can save a line here:

```swift
if let username = urlComponents?.user, let password = urlComponents?.password {
    return "\(username):\(password)"
}
return nil
```

Sources/KituraNet/URLParser.swift
```diff
@@ -22,29 +22,89 @@ import Foundation
...
+        }
+        set (newValue) {
+            self.urlComponents?.port = newValue
+        }
+    }
 
     /// The query parameters broken out.
-    public var queryParameters: [String: String] = [:]
+    public var queryParameters: [String: String] {
+        var _query: [String: String] = [:]
```
ianpartridge: This will calculate the `queryParameters` dictionary every time the property is accessed. For this property we should cache the dictionary.

nethraravindran: @ianpartridge I agree! Thanks for pointing this out :)

Sources/KituraNet/URLParser.swift
nethraravindran: Sure

Issue #154, opened 2019-01-19
regression: PR #143


pushkarnk:
Kitura tests fail with the changes related to #143 

We need to do the following:
1. Debug all Kitura failures and fix them
2. Subsequently measure performance and decide if the new changes could be acceptable

I've reverted #143 using #153

nethraravindran:
#143 caused failures in Kitura due to invalid url formed using incorrect components  `_urlComponents?.path = _urlString` 
In this `_urlString` is the uri string from `requestHead` containing both path and percent encoded query. Hence it has to be split and assigned to `path` and `percentEncodedQuery` of the URLComponents respectively.

nethraravindran:
Fix: https://github.com/IBM-Swift/Kitura-NIO/pull/156

pushkarnk:
Merged #156. Thanks @nethraravindran

PR #156, opened 2019-01-21
Build urlURL using URLComponents

nethraravindran:
This PR fixes all  Kitura failures reported in https://github.com/IBM-Swift/Kitura-NIO/issues/154

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -78,13 +78,13 @@ public class HTTPServerRequest: ServerRequest {
 
         _urlComponents?.port = Int(localAddressPort)
 
-        let uriComponents = _urlString.components(separatedBy: "?")
+        let uriComponents = _urlString.split(separator: "?")
         if uriComponents.count > 0 {
-            _urlComponents?.path = uriComponents[0]
+            _urlComponents?.path = String(uriComponents[0])
         }
 
         if uriComponents.count > 1 {
-            _urlComponents?.percentEncodedQuery = uriComponents[1]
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -67,22 +69,25 @@ public class HTTPServerRequest: ServerRequest {
...
             let hostname = localAddress
-            url.append(hostname == "127.0.0.1" ? "localhost" : hostname)
-            url.append(":")
-            url.append(String(describing: localAddressPort))
+            _urlComponents?.host = hostname == "127.0.0.1" ? "localhost" : hostname
+        }
+
+        _urlComponents?.port = Int(localAddressPort)
+
+        let uriComponents = _urlString.components(separatedBy: "?")
```
pushkarnk: It may be better to use `String.split()` here. See https://github.com/IBM-Swift/Kitura-net/pull/287

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -67,22 +69,25 @@ public class HTTPServerRequest: ServerRequest {
...
+        if uriComponents.count > 0 {
```
pushkarnk: This check may not be necessary.

pushkarnk: I'm sorry, this check will be needed if we use `String.split(separator:)`

pushkarnk:
I just rebased this on the latest master to make sure Kitura tests are run.

pushkarnk:
Kitura test have passed.

PR #165, opened 2019-02-04
add tests to test query parameters and redirection

pushkarnk:
Add two new tests to:
1. test query parameters, including values with % encoding
2. test redirection


pushkarnk:
Yes, the redirection test does test `ClientRequest`.

Regarding the query params test, though we are running the Kitura test suite on every pull request, we decided that every time a regression escapes Kitura-NIO/Kitura-net and is found by the Kitura tests, we'd add/port the corresponding test to Kitura-net/NIO. #154 was one such regression.

PR #166, opened 2019-02-05
Create Data using String.utf8 instead of NSString.data(using:)

pushkarnk:
We have `string.data(using:)` to create Data in a handful of places. This will be sub-optimal given that it implicitly goes through NSString, which is currently a known performance bottleneck in Swift 5. 

This is on the lines of https://github.com/IBM-Swift/Kitura/pull/1395

### DIFF
--- a/Sources/KituraNet/FastCGI/FastCGIServerRequest.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServerRequest.swift
@@ -55,7 +55,7 @@ public class FastCGIServerRequest: ServerRequest {
     /// The URL from the request in UTF-8 form
     /// This contains just the path and query parameters starting with '/'
     /// Use 'urlURL' for the full URL
-    public var url: Data { return requestUri?.data(using: .utf8) ?? Data() }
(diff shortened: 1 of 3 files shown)


Sources/KituraNet/FastCGI/FastCGIServerRequest.swift
```diff
@@ -55,7 +55,7 @@ public class FastCGIServerRequest: ServerRequest {
     /// The URL from the request in UTF-8 form
     /// This contains just the path and query parameters starting with '/'
     /// Use 'urlURL' for the full URL
-    public var url: Data { return requestUri?.data(using: .utf8) ?? Data() }
+    public var url: Data { return Data(urlURL.absoluteString.utf8) }
```
ianpartridge: Why was this changed to use `urlURL` rather than `requestUri`? The comment above says they are different.

pushkarnk: @ianpartridge Thanks for finding this. I've fixed it.

PR #168, opened 2019-02-06
tests: Rework .kitura-test to use Travis folding functions

djones6:
IBM-Swift/Package-Builder#161 added the ability to use Travis's output 'folding' to divide build logs into collapsible timed sections, making it much easier to navigate our build logs (particularly where we run the Kitura tests in addition to the Kitura-NIO tests).

This PR adds sections to the custom test script to produce these sections.  I've also reworked the error handling to ensure that test failures are surfaced (via return code) for backtrace generation.  Here's an example of this in action (with a deliberate crash inserted in the tests): https://travis-ci.org/djones6/Kitura-NIO/builds/489606985

A downside of this change is that `.kitura-test.sh` is now dependent on functions that are defined by Package-Builder, however, I don't think there's any intention of running it outside of this context.

### DIFF
--- a/.kitura-test.sh
+++ b/.kitura-test.sh
@@ -1,41 +1,44 @@
 # Run Kitura-NIO tests
+travis_start "swift_test"
+echo ">> Executing Kitura-NIO tests"
 swift test
-echo ">> Done running tests. Now preparing to run Kitura tests ..."
+SWIFT_TEST_STATUS=$?
+travis_end
+if [ $SWIFT_TEST_STATUS -ne 0 ]; then
+  echo ">> swift test command exited with $SWIFT_TEST_STATUS"
+  # Return a non-zero status so that Package-Builder will generate a backtrace
+  return $SWIFT_TEST_STATUS
+fi
 
 # Clone Kitura
-echo ">> cd .. && git clone https://github.com/IBM-Swift/Kitura && cd Kitura"
+set -e
+echo ">> Building Kitura"
+travis_start "swift_build_kitura"
 cd .. && git clone https://github.com/IBM-Swift/Kitura && cd Kitura
 
 # Set KITURA_NIO
-echo ">> export KITURA_NIO=1"
 export KITURA_NIO=1
 
 # Build once
-echo ">> swift build"
 swift build
 
 # Edit package Kitura-NIO to point to the current branch
-echo ">> swift package edit Kitura-NIO --path ../Kitura-NIO"
+echo ">> Editing Kitura package to use latest Kitura-NIO"
 swift package edit Kitura-NIO --path ../Kitura-NIO
-echo ">> swift package edit returned $?."
-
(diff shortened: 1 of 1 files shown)


.kitura-test.sh
```diff
@@ -1,41 +1,44 @@
...
-echo ">> swift package edit returned $?."
-
-# If the `swift package edit` command failed, exit with the same failure code
-PACKAGE_EDIT_RESULT=$?
-if [[ $PACKAGE_EDIT_RESULT != 0 ]]; then
-    echo ">> Failed to edit the Kitura-NIO dependency."
-    exit $PACKAGE_EDIT_RESULT
-fi
+travis_end
+set +e
```
pushkarnk: Why are we disabling the exit flag here?

.kitura-test.sh
```diff
@@ -1,41 +1,44 @@
...
+echo ">> Executing Kitura tests"
 swift test
-
-# If the tests failed, exit
-TEST_EXIT_CODE=$?
-if [[ $TEST_EXIT_CODE != 0 ]]; then
-    exit $TEST_EXIT_CODE
+SWIFT_TEST_STATUS=$?
+travis_end
+if [ $SWIFT_TEST_STATUS -ne 0 ]; then
```
pushkarnk: We could avoid this if we remove the `set +e` right?

.kitura-test.sh
```diff
@@ -1,41 +1,44 @@
...
+  echo ">> swift test command exited with $SWIFT_TEST_STATUS"
+  # Return a non-zero status so that Package-Builder will generate a backtrace
+  return $SWIFT_TEST_STATUS
+fi
 
 # Clone Kitura
-echo ">> cd .. && git clone https://github.com/IBM-Swift/Kitura && cd Kitura"
+set -e
```
pushkarnk: Why aren't we setting this at the start of the script?

djones6:
@pushkarnk to address your review comments regarding `set -e`, and also explain the absence of `set -x` (given that I originally suggested it in https://github.com/IBM-Swift/Kitura-net/pull/290):

I'd forgotten that in Package-Builder's `run_tests.sh`, we deliberately run the tests _without_ set -e, so that we can intercept a non-zero exit code (eg. from a crash) and try to generate a backtrace from a core.  This proves quite helpful in debugging intermittent failures,  but relies on the script not terminating immediately when the tests fail.  We do want the script to terminate if the Kitura clone + build fails, which is why I've included it around that section.

While developing https://github.com/IBM-Swift/Package-Builder/pull/161, I discovered that the Travis functions for producing markers in the output don't work correctly if running under `set -x` (the command is echoed, which causes the markers to appear twice, and breaks parsing).  For that reason, I've reintroduced some echo statements to indicate progress, but we don't need to be too verbose because now the output is nicely divided into sections.

PR #171, opened 2019-02-20
Add %-encoding in ClientRequest.init, remove in HTTPServerRequest.init

pushkarnk:
This is related to a test failure in CouchDB: https://github.com/IBM-Swift/Kitura-CouchDB/issues/91

During the initialisation of ClientRequest, the `path` component, which, in `ClientRequest`, includes `query` component, needs to be percent encoded. On the other side, during the initialisation of an HTTPServerRequest, we'd need to remove the percent encoding.

ClientRequest.Options.path includes the fragment, path and query components. The only character which is permissible in a fragment and a query but not in a path is `?`. To avoid further parsing here, we could go with percent-encoding a `?` in the path string too.

**A word on encoding `?`**
This does not pertain to the `?` as in: `https://66o.tech/path?q=v`. This pertains to a `?` as in: `https://66o.tech/pa?th?q=?v` which after %-encoding will become `https://66o.tech/pa%3Fth?q=%3Fv`, though `https://66o.tech/pa%3Fth?q=?v` is permissible per the [definition of NSCharacterSet.urlQueryAllowed](https://github.com/apple/swift-corelibs-foundation/blob/b23dc8e249987442811e9e0097c326d7083784ac/CoreFoundation/URL.subproj/CFURLComponents_URIParser.c#L36).

In the proposed change, the `?` character will not be percent-encoded in the path (fragment, path and query), though it should be encoded in the "path". I'm assuming that paths like "/this?path" are rare in occurrence. Percent encoding '?' will make parsing ClientRequest.path all the more complicated and we may want to avoid it until it is explicitly asked for. 

The fact that `ClientRequest.Options.path` includes path and query is a design weakness in this context.

Other components like `host` and `user` may also be percent-encoded, but are those cases really common?

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -126,7 +126,18 @@ public class HTTPServerRequest: ServerRequest {
 
     private var enableSSL: Bool = false
 
-    private var _urlString: String
+    private var rawURLString: String
+
+    private var urlStringPercentEncodingRemoved: String?
+
+    private var _urlString: String {
+        guard let urlStringPercentEncodingRemoved = self.urlStringPercentEncodingRemoved else {
+            let _urlStringPercentEncodingRemoved = rawURLString.removingPercentEncoding ?? rawURLString
+            self.urlStringPercentEncodingRemoved = _urlStringPercentEncodingRemoved
+            return _urlStringPercentEncodingRemoved
+        }
+        return urlStringPercentEncodingRemoved
+    }
 
     private static func host(socketAddress: SocketAddress?) -> String {
         guard let socketAddress = socketAddress else {
@@ -148,7 +159,7 @@ public class HTTPServerRequest: ServerRequest {
         self.method = requestHead.method.string()
         self.httpVersionMajor = requestHead.version.major
         self.httpVersionMinor = requestHead.version.minor
-        self._urlString = requestHead.uri
+        self.rawURLString = requestHead.uri
         self.enableSSL = enableSSL
     }
 
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -238,7 +238,7 @@ public class ClientRequest {
                 if thePath.first != "/" {
                     thePath = "/" + thePath
                 }
-                path = thePath
+                path = thePath.addingPercentEncoding(withAllowedCharacters: NSCharacterSet.urlQueryAllowed) ?? thePath
                 self.path = path
             case .username(let userName):
                 self.userName = userName
--- a/Tests/KituraNetTests/ClientRequestTests.swift
+++ b/Tests/KituraNetTests/ClientRequestTests.swift
@@ -98,6 +98,20 @@ class ClientRequestTests: KituraNetTest {
         let options = ClientRequest.parse("https://username:password@66o.tech:8080/path?key=value")
         let testRequest = ClientRequest(options: options, callback: testCallback)
         XCTAssertEqual(testRequest.url, "https://username:password@66o.tech:8080/path?key=value")
+
+        let options1: [ClientRequest.Options] = [ .schema("https"),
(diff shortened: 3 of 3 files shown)


pushkarnk:
@djones6 Could you please review? There are three caveats here:

1. `ClientRequest.Options.path` includes path and query. Query allows `?`, path doesn't. If we decide to %-encode `?`, parsing the `ClientRequest.Options.path` to get the path and query might get challenging. If we do not encode `?` we cannot support paths like `/with?question`, which I assume are rare. Is this acceptable for now?

2. The `host` and `user` components may also have unsupported characters. Do you think I must include their %-encoding in this PR?

3. Finally, adding and removing %-encoding will hit performance with Swift 5.

djones6:
@pushkarnk How does ClientRequest in Kitura-net handle this? (is it handled for us by curl?)
I couldn't find code (or tests) in Kitura-net relating to this, but I'm presuming it must be handled somehow given that the CouchDB tests are passing.

To answer your questions above:
1) I think it's acceptable for now to not support embedding `?` within the path. We should support it if/when ClientRequest is redesigned to have a separate path and query component.

2) It seems very unlikely that a hostname would include disallowed characters.  Maybe a username might, but the field that seems absolutely likely to contain weird characters is `password`.  I can't see any special handling of that field currently.

Another question would be: should we instead by transmitting the credentials in an `Authorization` header? I gather that url encoded authorization is strongly discouraged, for reasons such as inadvertent disclosure / leakage eg. if the URL is logged or stored somewhere along the way.  Doing so would avoid the need to url encode (as the username:password value is base64 encoded).

3) I don't think there's a way to avoid the performance hit: escaping is fairly costly, but it's required for correctness, _unless_ we require the user to escape values.  (We also must be careful not to double-escape).  In terms of the amount of performance cost, the whole process of making an HTTP request and parsing a response is already a fairly expensive operation, so I suspect the cost of escaping the query string won't be significant.

pushkarnk:
Thank for the review @djones6 ... To answer the two questions you asked:
1. Yes, with Kitura-net, libcurl handles adding and dropping %-encoding.

2. Yes, if username/password are supplied as `ClientRequest.Options` we could create an `authorisation` header instead of encoding them in the URL. And we'll have to do that on Kitura-net too. But it is going to be a breaking change (as in, users will start seeing a different value for `ClientRequest.url`).

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -126,7 +126,16 @@ public class HTTPServerRequest: ServerRequest {
...
-    private var _urlString: String
+    private var rawURLString: String
+
+    private var urlStringPercentEncodingRemoved: String?
+
+    private var _urlString: String {
+        if urlStringPercentEncodingRemoved == nil {
+            urlStringPercentEncodingRemoved = rawURLString.removingPercentEncoding ?? rawURLString
+        }
+        return urlStringPercentEncodingRemoved!
```
djones6: I can see this code is correct, but could we use guard let here so that we don't need the force-unwrap?

pushkarnk: @djones6 Using `guard-let` and not force-unwrapping it can look only as good as this:
```
    private var _urlString: String {
        guard let urlStringPercentEncodingRemoved == self.urlStringPercentEncodingRemoved else {
            urlStringPercentEncodingRemoved = rawURLString.removingPercentEncoding ?? rawURLString
            return rawURLString.removingPercentEncoding ?? rawURLString
        }   
        return urlStringPercentEncodingRemoved
    }
```

djones6:
```swift
            let result = rawURLString.removingPercentEncoding ?? rawURLString
            urlStringPercentEncodingRemoved = result
            return result
```
It's important not to run `removingPercentEncoding` twice as it is quite expensive.

pushkarnk: Done. Thanks @djones6 !

PR #175, opened 2019-03-11
HTTPServerResponse can hold a strong Channel reference

pushkarnk:
HTTPServerResponse currently has a weak reference to the Channel over which the response will be written. Since a retain cycle between Channel and HTTPServerResponse isn't likely, it makes sense to keep a strong reference to the Channel. This helps with an edge case seen with TechEmpower where the Channel gets 'prematurely' cleaned-up before the last response could be written on a longstanding connection. Keeping a strong reference also eliminates some nil checks, making the code easier to read.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -24,7 +24,7 @@ import Foundation
 public class HTTPServerResponse: ServerResponse {
 
     /// The channel to which the HTTP response should be written
-    private weak var channel: Channel?
+    private let channel: Channel
 
     /// The handler that processed the HTTP request
     private weak var handler: HTTPRequestHandler?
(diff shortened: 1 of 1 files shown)


pushkarnk:
We first [introduced a weak reference to ChannelHandlerContext](https://github.com/IBM-Swift/Kitura-NIO/pull/48) to avoid a retain cycle. Later we replaced it with a `Channel` instead. Though I don't think a retain cycle `HTTPServerResponse <---> Channel` is possible, a weak reference serves as a nice signalling mechanism to indicate that the underlying socket has been closed and there is no value in writing data to it.

I am convinced that the change proposed in this pull request is not the right way of solving this apparent problem. Closing this PR for now.

pushkarnk:
After some discussing, I created #186 which seems to be a better way to avoid this problem, and with lesser changes.

PR #176, opened 2019-03-12
Rebase Kitura-NIO on NIO2

pushkarnk:
This patch is to adopt the NIO 2 API. The change include adopting API changes in NIO and a seamless movement from NIOOpenSSL to NIOSSL.


pushkarnk:
Disabling code coverage tests for now. They fail with a "Swift 5.0 unsupported" message.

pushkarnk:
Disabled Kitura testing in Travis due to this: https://bugs.swift.org/browse/SR-10100

Package.swift
```diff
@@ -28,8 +28,9 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(url: "https://github.com/apple/swift-nio.git", from: "1.0.0"),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "1.0.1"),
+        // TODO: Update after swift-nio and swift-nio-ssl are tagged
+        .package(url: "https://github.com/apple/swift-nio.git", .branch("master")),
```
weissi: I think you should put `from: "2.0.0-convergence.1"` here instead of `.branch("master")`

pushkarnk: Thanks @weissi !

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -158,19 +158,19 @@ public class HTTPServer: Server {
...
 
     /// URI for which the latest WebSocket upgrade was requested by a client
     var latestWebSocketURI: String?
 
     /// Determines if the request should be upgraded and adds additional upgrade headers to the request
-    private func shouldUpgradeToWebSocket(webSocketHandlerFactory: ProtocolHandlerFactory, head: HTTPRequestHead) -> HTTPHeaders? {
+    private func shouldUpgradeToWebSocket(channel: Channel, webSocketHandlerFactory: ProtocolHandlerFactory, head: HTTPRequestHead) -> EventLoopFuture<HTTPHeaders?> {
         self.latestWebSocketURI = head.uri
-        guard webSocketHandlerFactory.isServiceRegistered(at: head.uri) else { return nil }
+        guard webSocketHandlerFactory.isServiceRegistered(at: head.uri) else { return channel.eventLoop.makeSucceededFuture(nil) }
```
ianpartridge: Why is this future succeeded? If this is an error case shouldn't it be failed?

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -118,7 +118,7 @@ public class HTTPServerRequest: ServerRequest {
         _urlComponents?.scheme = self.enableSSL ? "https" : "http"
 
         var localAddress = ""
-        var localAddressPort = UInt16(0)
+        var localAddressPort = Int(0)
```
ianpartridge: I think `Int` is automatically inferred so just `var localAddressPort = 0`?

.travis.yml
```diff
@@ -15,43 +15,25 @@ matrix:
...
-      osx_image: xcode9.2
-      sudo: required
-      env: SWIFT_SNAPSHOT=4.0.3 BREW_INSTALL_PACKAGES="libressl" CUSTOM_TEST_SCRIPT=.kitura-test.sh
-    - os: osx
-      osx_image: xcode9.4
-      sudo: required
-      env: SWIFT_SNAPSHOT=4.1.2 JAZZY_ELIGIBLE=true CODECOV_ELIGIBLE=true BREW_INSTALL_PACKAGES="libressl"
+      env: DOCKER_IMAGE=ubuntu:18.04 SWIFT_SNAPSHOT=$SWIFT_DEVELOPMENT_SNAPSHOT DOCKER_PRIVILEGED=true
     - os: osx
       osx_image: xcode10.1
```
djones6: Needs to be xcode10.2, and we should be dropping the SWIFT_SNAPSHOT

Sources/KituraNet/HTTP/HTTPServer.swift
pushkarnk: @ianpartridge Yes, it does look wrong. But a failed future needs an Error value to be passed. Here, we're just refusing an upgrade to WebSocket because a service wasn't registered at the requested path. There is no error type defined for this. In the NIO 1 API, the return type of this function `shouldUpgrade` was `HTTPHeaders?`, and we returned a nil to indicate an upgrade refusal. Given this background, even if we had an error type I'm not sure how the WebSocket upgrader would respond to a failed future with an error (I'll take a look at this).

pushkarnk: From a superficial look at the [WebSocketUpgrader](https://github.com/apple/swift-nio/blob/master/Sources/NIOWebSocket/WebSocketUpgrader.swift#L127) I don't think the upgrader is equipped to handle a failed future with an error(I might be wrong). I propose we take this up through an issue later.

ianpartridge: OK that's fine.

PR #181, opened 2019-03-15
Close channels if errors reach the last handler

pushkarnk:
Motivation

This is related to issue #173 If an HTTP error is encountered in the HTTPServer, it is essential that we close the connection after a possible handling of the error. As the issue describes it, this is crucial in the error reaches the last handler of the pipeline and none of the previous handlers were able to handle it. `HTTPRequestHandler` and `HTTPClientHandler` are final handlers of an HTTP server pipeline and an HTTP client pipeline respectively. `HTTPRequestHandler` does handle some errors. We make sure the connection is closed only after the error response is sent back. `HTTPClientHandler` does not do any error handling, so we just close the connection there.

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -653,6 +653,7 @@ class HTTPClientHandler: ChannelInboundHandler {
 
      func errorCaught(ctx: ChannelHandlerContext, error: Error) {
          // No errors to handle, simply close the channel
+         Log.error("ClientRequest: Error \(error) was received. The connection will be closed because we are neither handling this error nor can it be propagated.")
          ctx.close(promise: nil)
      }
 }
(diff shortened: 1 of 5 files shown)


Sources/KituraNet/ClientRequest.swift
```diff
@@ -650,4 +650,9 @@ class HTTPClientHandler: ChannelInboundHandler {
             }
          }
      }
+
+     func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+         // No errors to handle, simply close the channel
```
weissi: @pushkarnk this looks great. You might want to consider logging the errors when you close? But that's up to you.

pushkarnk: Yep, makes sense to log an error. I've done that. Thanks @weissi !

pushkarnk:
Thanks for the review @weissi

PR #182, opened 2019-03-19
ci: Update to Swift 4.2.3, use official Docker images

djones6:
Updates the CI to use the latest Swift 4.2.3 release, and the official Swift Docker images.
Using the official Docker images can speed up CI build times as the build does not need manually installing from swift.org, and system dependencies are already set up.


pushkarnk:
@djones Are we not going to test on Ubuntu 14.04 anymore?

djones6:
My argument is that, as 14.04 goes out of LTS in a few weeks time, and will then only be supported via (paid) extended maintenance, Kitura users should already have moved off it.  The official Swift Docker images are based on 16.04, and there's [discussion](https://forums.swift.org/t/ubuntu-14-04/17859/15) of not releasing Swift 5 on 14.04 (although to date, builds are still being produced).

pushkarnk:
Thanks for the clarification @djones6

PR #187, opened 2019-03-22
Basic support for unix domain sockets

pushkarnk:
This is a replication of the [unix domain socket support on Kitura-net](https://github.com/djones6/Kitura-net/commit/357806e0b491f6afcae235644a6e0d0fa8b31e24). This pull request proposes an additive change to the `HTTP.request()` and `ClientRequest.init(options:)` APIs via the addition of a new optional argument named `socketPath`, which defaults to nil to ensure zero breakage of ClientRequest. That's about the client side support. For the sake of testing, we also have an internal `listen(unixDomainSocket:)` that provides server-side support for listening on unix domain sockets. One of the existing Kitura-NIO tests `testLargePosts` is modified to work with a unix socket.


Tests/KituraNetTests/LargePayloadTests.swift
```diff
@@ -39,10 +39,25 @@ class LargePayloadTests: KituraNetTest {
 
     private let delegate = TestServerDelegate()
 
+    private func uniqueTemporaryFilePath() -> String {
```
djones6: This all looks fine, I used a slightly different implementation in the equivalent Kitura-net PR - take a look at my `setUp()` and `tearDown()` in [`Tests/KituraNetTests/UnixSocketTests.swift`](https://github.com/IBM-Swift/Kitura-net/blob/aa1755c6d88ae8dc4b9483a0dc608e95251816d5/Tests/KituraNetTests/UnixSocketTests.swift)
\- in particular, it would be nice to not leave a bunch of temporary files lying around after running the tests.

Also, is it possible for the test to assert that the client is connecting over a Unix socket? I did this in the test I referred to above, but I'm not sure if this is feasible with NIO.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -48,8 +48,8 @@ public class HTTP {
         return ClientRequest(url: url, callback: callback)
     }
 
-    public static func request(_ options: [ClientRequest.Options], callback: @escaping ClientRequest.Callback) -> ClientRequest {
-        return ClientRequest(options: options, callback: callback)
+    public static func request(_ options: [ClientRequest.Options], socketPath: String? = nil, callback: @escaping ClientRequest.Callback) -> ClientRequest {
```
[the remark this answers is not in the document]
pushkarnk: A recently merged PR has copied all of the documentation comments from Kitura-net to Kitura-NIO :)

pushkarnk:
Thanks for the review @djones6 I've addressed all of your comments.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -276,14 +299,23 @@ public class HTTPServer: Server {
...
             self.lifecycleListener.performFailCallbacks(with: error)
-            Log.error("Error trying to bind to \(port): \(error)")
+            switch socket {
+            case .tcp(let port):
+                Log.error("Error trying to bind to \(port): \(error)")
+            case .unix(let socketPath):
+                Log.error("Error trying to bind to \(socketPath): \(error)")
+            }
             throw error
         }
```
djones6: The line below this one:
```swift
Log.info("Listening on port \(self.port!)")
```
blows up if you listen on a Unix socket and have logging enabled...

djones6: I pushed a fix for this

djones6:
@pushkarnk I pushed https://github.com/IBM-Swift/Kitura-NIO/pull/187/commits/f04d4eb2e07d469b222bcaf07d19ee61997eded7 because it is otherwise not possible to start, stop and then restart the server.  This also makes it consistent with BlueSocket, which removes the file at the socket path if it already exists.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -275,20 +298,33 @@ public class HTTPServer: Server {
             }
             .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
 
+        let listenerDescription: String
         do {
-            serverChannel = try bootstrap.bind(host: "0.0.0.0", port: port).wait()
+            switch socket {
+            case SocketType.tcp(let port):
+                listenerDescription = "port \(port)"
```
djones6: I just realized this isn't quite right: this will result in `port 0` when listening on an ephemeral port.

pushkarnk: I've fixed this in https://github.com/IBM-Swift/Kitura-NIO/pull/187/commits/c7154c8fb7af29c13f8f9db4c4d417cca4187efc

pushkarnk:
@djones6 Redirection creates new `ClientRequest` objects. If an HTTP server listening on a unix domain socket returns a redirection request (302), the new ClientRequest must also connect to the same unix domain socket (if the value of the `Location` header is just another `path` i.e starts with `/`). If the value of the `Location` header is a full URL, we cannot make assumptions about it being hosted on the same unix domain socket. 

This commit handles the case of the `Location` having just a different path https://github.com/IBM-Swift/Kitura-NIO/pull/187/commits/41062b60b0a88f795bfb6f1932f16231107adf8c

With this the Kitura tests must pass.

PR #194, opened 2019-04-11
[kitura-nio-1.0] Cherry-pick PR#193

pushkarnk:
Sometimes the value of the Host header could be of the type
"localhost:8080". We've been using URLComponents to fabricate the
URL pertaining to an HTTPServerRequest. In cases where, the Host
header includes the port number, we'll have to pick up only the
host name and ignore the port number.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,11 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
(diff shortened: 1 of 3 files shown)


pushkarnk:
Disabling Kitura tests until we have a decision on https://github.com/IBM-Swift/Kitura/pull/1440

PR #205, opened 2019-05-16
Allow Kitura-NIO to only listen on one network address

TaborKelly:
## Description
@billabt was nice enough to add support for this at the BlueSocket layer [here](https://github.com/IBM-Swift/BlueSocket/commit/9d77d6b2126d099f5e56cace9727068f145daaec). However, I really want to use that functionality to fix https://github.com/IBM-Swift/Kitura/issues/1450.

I also have code ready to go for https://github.com/IBM-Swift/Kitura-net/pull/299 and Kitura.

Obviously, this is the Kitura-NIO version of the Kitura-Net change.

## Motivation and Context
This is necessary to close https://github.com/IBM-Swift/Kitura/issues/1450. Should I open another issue for Kitura-NIO?

## How Has This Been Tested?
Manual testing by myself. I also modified the test harness to use the new code path.

## Checklist:

- [x] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)
- [x] If applicable, I have updated the documentation accordingly.
- [x] If applicable, I have added tests to cover my changes.

### DIFF
--- a/Sources/KituraNet/FastCGI/FastCGIServer.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServer.swift
@@ -59,7 +59,7 @@ public class FastCGIServer: Server {
     /// - Parameter on: port number for new connections
     /// - Parameter address: The address of the network interface to listen on. Defaults to nil, which means this
     ///             server will listen on all interfaces.
-    public func listen(on port: Int, address: String?) throws {
+    public func listen(on port: Int, address: String? = nil) throws {
         fatalError("FastCGI is not implemented yet.")
     }
 
@@ -71,7 +71,7 @@ public class FastCGIServer: Server {
     /// - Parameter delegate: the delegate handler for FastCGI/HTTP connections
     ///
     /// - Returns: a new `FastCGIServer` instance
-    public static func listen(on port: Int, address: String?, delegate: ServerDelegate?) throws -> FastCGIServer {
+    public static func listen(on port: Int, address: String? = nil, delegate: ServerDelegate?) throws -> FastCGIServer {
         fatalError("FastCGI not implemented yet.")
     }
 
--- a/Sources/KituraNet/FastCGI/FastCGIServer.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServer.swift
@@ -29,6 +29,9 @@ public class FastCGIServer: Server {
     /// Port number for listening for new connections
     public private(set) var port: Int?
 
+    /// Has the same meaning as in `getaddrinfo()`.
+    public private(set) var node: String?
+
     /// A server state.
     public private(set) var state: ServerState = .unknown
 
@@ -53,8 +56,28 @@ public class FastCGIServer: Server {
     /// Listens for connections on a socket
     ///
     /// - Parameter on: port number for new connections
+    /// - Parameter node: has the same meaning as in `getaddrinfo()`
+    public func listen(on port: Int, node: String?) throws {
+        fatalError("FastCGI is not implemented yet.")
+    }
+
+    /// Listens for connections on a socket
+    ///
+    /// - Parameter on: port number for new connections
+    @available(*, deprecated, message: "use 'listen(on:node) throws' instead")
     public func listen(on port: Int) throws {
         fatalError("FastCGI is not implemented yet.")
+    }
+
+    /// Static method to create a new `FastCGIServer` and have it listen for conenctions
+    ///
+    /// - Parameter on: port number for accepting new connections
+    /// - Parameter node: has the same meaning as in `getaddrinfo()`
+    /// - Parameter delegate: the delegate handler for FastCGI/HTTP connections
+    ///
+    /// - Returns: a new `FastCGIServer` instance
+    public static func listen(on port: Int, node: String?, delegate: ServerDelegate?) throws -> FastCGIServer {
+        fatalError("FastCGI not implemented yet.")
     }
 
     /// Static method to create a new `FastCGIServer` and have it listen for conenctions
@@ -63,6 +86,7 @@ public class FastCGIServer: Server {
     /// - Parameter delegate: the delegate handler for FastCGI/HTTP connections
     ///
     /// - Returns: a new `FastCGIServer` instance
+    @available(*, deprecated, message: "use 'listen(on:node:delegate) throws' instead")
     public static func listen(on port: Int, delegate: ServerDelegate?) throws -> FastCGIServer {
         fatalError("FastCGI not implemented yet.")
     }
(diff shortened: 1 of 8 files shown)


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -258,14 +261,21 @@ public class HTTPServer: Server {
...
+     - Parameter node: has the same meaning as in `getaddrinfo()`
+     */
+    public func listen(on port: Int, node: String?) throws {
         self.port = port
-        try listen(.tcp(port))
+        self.node = node
+        try listen(.tcp(port, node))
+    }
+
+    @available(*, deprecated, message: "use 'listen(on:node) throws' instead")
```
pushkarnk: I'm not sure if we should deprecate `listen(on:)`. We could make this a simple additive change by introducing a new `listen(on:node)` or `listen(on:address)`

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -360,21 +370,27 @@ public class HTTPServer: Server {
...
      ````
 
      - Parameter on: Port number for accepting new connections.
+     - Parameter node: has the same meaning as in `getaddrinfo()`
      - Parameter delegate: The delegate handler for HTTP connections.
 
      - Returns: A new instance of a `HTTPServer`.
     */
-    public static func listen(on port: Int, delegate: ServerDelegate?) throws -> ServerType {
+    public static func listen(on port: Int, node: String?, delegate: ServerDelegate?) throws -> ServerType {
```
pushkarnk: Can we have the `node`/`address` parameter have a default value of `nil`? To make sure this is not a breaking change.

Sources/KituraNet/Server/Server.swift
```diff
@@ -26,20 +26,40 @@ public protocol Server {
     /// Port number for listening for new connections.
     var port: Int? { get }
 
+    /// Has the same meaning as in `getaddrinfo()`.
+    var node: String? { get }
```
pushkarnk: Can we call this `address`?

Sources/KituraNet/Server/Server.swift
```diff
@@ -26,20 +26,40 @@ public protocol Server {
...
+
     /// A server state.
     var state: ServerState { get }
 
     /// Listen for connections on a socket.
     ///
     /// - Parameter on: port number for new connections (eg. 8080)
+    /// - Parameter node: has the same meaning as in `getaddrinfo()`
+    func listen(on port: Int, node: String?) throws
```
pushkarnk: Repeat: `node` can have a default value of `nil`

Sources/KituraNet/Server/Server.swift
```diff
@@ -26,20 +26,40 @@ public protocol Server {
...
+
+    /// Listen for connections on a socket.
+    ///
+    /// - Parameter on: port number for new connections (eg. 8080)
+    @available(*, deprecated, message: "use 'listen(on:node) throws' with instead")
```
pushkarnk: repeat: can we avoid deprecation here?

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -58,6 +58,9 @@ public class HTTPServer: Server {
     /// The TCP port on which this server listens for new connections. If `nil`, this server does not listen on a TCP socket.
     public private(set) var port: Int?
 
+    /// Has the same meaning as in `getaddrinfo()`.
+    public private(set) var node: String?
```
pushkarnk: Can we call this `address` please?

TaborKelly:
I believe that my latest changes fix everything that was requested and bring this PR in line with the one for Kitura-Net. Let me know if I missed anything.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -258,14 +262,17 @@ public class HTTPServer: Server {
...
+     try server.listen(on: 8080, address: "localhost")
      ````
 
      - Parameter on: Port number for new connections, e.g. 8080
-    */
-    public func listen(on port: Int) throws {
+     - Parameter address: The address of the network interface to listen on. Defaults to nil, which means this server
+                          will listen on all interfaces.
+     */
+    public func listen(on port: Int, address: String?) throws {
```
pushkarnk: This must match the Kitura-net version: 
`public func listen(on port: Int, address: String? = nil) throws {`

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -360,18 +367,20 @@ public class HTTPServer: Server {
...
 
      - Parameter on: Port number for accepting new connections.
+     - Parameter address: The address of the network interface to listen on. Defaults to nil, which means this server
+                 will listen on all interfaces.
      - Parameter delegate: The delegate handler for HTTP connections.
 
      - Returns: A new instance of a `HTTPServer`.
     */
-    public static func listen(on port: Int, delegate: ServerDelegate?) throws -> ServerType {
+    public static func listen(on port: Int, address: String?, delegate: ServerDelegate?) throws -> ServerType {
```
pushkarnk: Same feedback here. This must match Kitura-net:
`public static func listen(on port: Int, address: String? = nil, delegate: ServerDelegate?) throws -> HTTPServer`

Sources/KituraNet/FastCGI/FastCGIServer.swift
```diff
@@ -53,17 +57,21 @@ public class FastCGIServer: Server {
     /// Listens for connections on a socket
     ///
     /// - Parameter on: port number for new connections
-    public func listen(on port: Int) throws {
+    /// - Parameter address: The address of the network interface to listen on. Defaults to nil, which means this
+    ///             server will listen on all interfaces.
+    public func listen(on port: Int, address: String?) throws {
```
djones6: Although FastCGI is not implemented yet, the function signature should still be consistent with Kitura-net (as per @pushkarnk 's comments on HTTPServer) - need to add the default `nil` value.

Sources/KituraNet/FastCGI/FastCGIServer.swift
```diff
@@ -53,17 +57,21 @@ public class FastCGIServer: Server {
...
     /// Static method to create a new `FastCGIServer` and have it listen for conenctions
     ///
     /// - Parameter on: port number for accepting new connections
+    /// - Parameter address: The address of the network interface to listen on. Defaults to nil, which means this
+    ///             server will listen on all interfaces.
     /// - Parameter delegate: the delegate handler for FastCGI/HTTP connections
     ///
     /// - Returns: a new `FastCGIServer` instance
-    public static func listen(on port: Int, delegate: ServerDelegate?) throws -> FastCGIServer {
+    public static func listen(on port: Int, address: String?, delegate: ServerDelegate?) throws -> FastCGIServer {
```
djones6: (same here)

TaborKelly:
Good catch. I missed that since they conformed to the same Server protocol.

Issue #178, opened 2019-03-13
ClientRequest auth should use an Authorization header


djones6:
In https://github.com/IBM-Swift/Kitura-NIO/pull/171 we noted that currently, the username and password fields are being sent as part of the URL for basic authentication, and that they are currently not percent-escaped by the NIO client.

Rather than escaping them, I believe they should instead be sent in [an `Authorization` header](https://en.wikipedia.org/wiki/Basic_access_authentication). This would alleviate the need to escape their values, besides being recommended practice.  Normally the Authorization header is sent in response to a server sending a `WWW-Authenticate` header, but I believe it's valid to provide it outside of that flow.

This would be a behavioural (breaking?) change as the format of the URL would change where basic authentication is being used, and would require recipients to support parsing the Authorization header (which I believe they should do, but I can imagine there may be implementations out there that assumes they'll be tacked on to the URL).

Both Kitura-NIO and Kitura-net would need to be changed to behave consistently.

djones6:
We discovered that for Kitura-net, `libcurl` already handles this for us: credentials prepended to the URL are removed and sent instead as an `Authorization` header.

Need to confirm whether the same happens for Kitura-NIO.

djones6:
This was merged (for Kitura-NIO).

PR #208, opened 2019-06-14
Send Basic Auth credentials in an Authorization header

RudraniW:
The username and password are currently sent as a part of the url. This can be a challenge
if the url need to be percent encoded. The username and password also need to be percent
encoded then. It is easier to send username and password in the Authorization header. Note that
this will change the behaviour of `ClientRequent.url`.

This is for #178 . **It is a breaking change.**

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -269,12 +269,12 @@ public class ClientRequest {
         }
         self.path = fullPath
 
-        if let username = url.user {
-            self.userName = username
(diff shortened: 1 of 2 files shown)


Sources/KituraNet/ClientRequest.swift
```diff
@@ -269,12 +269,12 @@ public class ClientRequest {
         }
         self.path = fullPath
 
-        if let username = url.user {
```
pushkarnk: We must retain this code. The username and password will continue to be accessible using the `userName` and `password` properties. They only won't be included in the `url` property.

PR #213, opened 2019-07-11
Add default Accept header and port number to Host

harish1992:
Kitura-net adds a default Accept header and appends port number to host name if
non-default ports(443/80) are used. In order to maintain parity default Accept
header is added and port name is appended if non default port is used.


harish1992:
#### The changes have been tested with:
- Kitura-WebSocket-NIO
- Kitura-Sample
- Kitura-CouchDB
- SwiftMetrics

PR #214, opened 2019-07-12
Fix to set property ClientResponse.status right

harish1992:
Kitura-NIO wrongly returned -1 as status code.The property
ClientResponse.status now updates when ClientResponse.statusCode is
updated.Observer didSet is removed from ClientResponse.status as observers of
both status and statusCode trigger each other and cause stackoverflow.The
changes made ensure the property ClientResonse.status is set right.


harish1992:
#### The changes have been tested with:
- Kitura-WebSocket-NIO
- Kitura-Sample
- Kitura-CouchDB
- SwiftMetrics

pushkarnk:
Thanks.

PR #215, opened 2019-07-15
Allow HTTPServerRequest creation on the current EventLoop only

pushkarnk:
An `HTTPServerRequest` abstracts a received HTTP request. Though an
`HTTPServerRequest` must be strictly associated with an HTTP connection,
Kitura-NIO allowed creation of HTTPServerRequests from anywhere. Consequently,
to build the `URL` for the `urlURL` property, when we need to access
`ChannelHandlerContext` properties like `localAddress` or `remoteAddress`, that
code needs to be dispatched to the EventLoop. Because the computation of
`urlURL` is synchronous, we also need to wait() for the EventLoop to return. If
`urlURL` were invoked on an EventLoop, like we do in `Kitura-WebSocket-NIO`, we
fail with a SwiftNIO assertion that doesn't permit wait() to be invoked on an
`EventLoop`.

The best possible solution to this issue is to restrict the HTTPServerRequest
creation on the EventLoop associated with the underlying HTTP connection only.
The `ChannelHandlerContext` properties could then be accessed in the 
initializer.

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -194,18 +194,21 @@ public class HTTPServer: Server {
 
     /// Creates upgrade request and adds WebSocket handler to pipeline
     private func upgradeHandler(webSocketHandlerFactory: ProtocolHandlerFactory, request: HTTPRequestHead) -> EventLoopFuture<Void> {
-        guard let ctx = self.ctx else { fatalError("Cannot create ServerRequest") }
-        let serverRequest = HTTPServerRequest(ctx: ctx, requestHead: request, enableSSL: false)
-        let websocketConnectionHandler = webSocketHandlerFactory.handler(for: serverRequest)
-        let future = ctx.channel.pipeline.addHandler(websocketConnectionHandler)
-        if let _extensions = request.headers["Sec-WebSocket-Extensions"].first {
-            for handler in webSocketHandlerFactory.extensionHandlers(header: _extensions) {
-                _ = future.flatMap {
-                    ctx.channel.pipeline.addHandler(handler, position: .before(websocketConnectionHandler))
+        guard let ctx = self.ctx else { fatalError("The channel was probably closed during a protocol upgrade.") }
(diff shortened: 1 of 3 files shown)


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -194,18 +194,21 @@ public class HTTPServer: Server {
...
-                _ = future.flatMap {
-                    ctx.channel.pipeline.addHandler(handler, position: .before(websocketConnectionHandler))
+        guard let ctx = self.ctx else { fatalError("The channel was probably closed during a protocol upgrade.") }
+        return ctx.eventLoop.submit {
+            let request = HTTPServerRequest(ctx: ctx, requestHead: request, enableSSL: false)
+            return webSocketHandlerFactory.handler(for: request)
+        }.flatMap { (webSocketConnectionHandler: ChannelHandler) -> EventLoopFuture<Void> in
+            var future = ctx.channel.pipeline.addHandler(webSocketConnectionHandler)
+            if let _extensions = request.headers["Sec-WebSocket-Extensions"].first {
+                for handler in webSocketHandlerFactory.extensionHandlers(header: _extensions) {
```
weissi: you could also do `ctx.channel.pipeline.addHandlers(webSocketHandlerFactory.extensionHandlers(header: _extensions), positition: .before(...))`. With `addHandlers` you can add multiple handlers in one go.

pushkarnk: thanks @weissi !

pushkarnk:
Testing todos:

- [ ] Kitura-WebSocket-NIO
- [ ] WebSocket compression tests
- [ ] Kitura-Sample
- [ ] Kitura-CouchDB
- [ ] SwiftMetrics

pushkarnk:
> Testing todos:
> 
> * [x]  Kitura-WebSocket-NIO
> * [x]  WebSocket compression tests
> * [x]  Kitura-Sample
> * [x]  Kitura-CouchDB
> * [x]  SwiftMetrics

The fix has come out clean from these.

Issue #206, opened 2019-05-22
Ability to limit incoming request size (Kitura#1384)


pushkarnk:
Copied from https://github.com/IBM-Swift/Kitura/issues/1384

Kitura currently does not have the concept of a limit on incoming request size. It is possible to destabilize a server by sending (accidentally or maliciously) a large request payload which either impairs performance, or exhausts available memory.

A limit should be configurable on a server which causes requests above that size to be rejected, and the connection closed.

weissi:
Ouch! Instead of limiting the incoming request size (which usually causes trouble when doing file uploads (ask Vapor for example)), I'd suggest implementing streaming with back-pressure. That way the maximum amount of memory can remain constant and yet you can support arbitrarily large file uploads. Of course a user of Kitura should be able to control what's too much. So in some JSON API, maybe 10 MB might be a maximum but for another route that accepts file uploads there shouldn't be a maximum because files can be streamed straight to disk (possibly policing the quota for the logged in user of course).

RudraniW:
- I have implemented the function to calculate size of the header. 
- I have implemented the function which will send the status code and status description when request size exceeds the default maximum request size.
- Now I am configuring the HTTP server to set maximum request size and connections.

pushkarnk:
@djones6 We did consider writing a new channel handler to track the number of raw bytes flowing into a channel. However, if that particular connection is supporting pipelining I don't think there's any way to mark HTTP request boundaries on a raw byte stream. Data pertaining to multiple HTTP requests could be flowing in sequentially. The HTTP message boundaries are drawn only after the HTTPDecoder converts a chunk of bytes to an `HTTPRequestHead` and a body (`ByteBuffer`). So, on such a connection how does one calculate the size of one HTTP message without having the an HTTP decoder parse the raw bytes ?

pushkarnk:
> Ouch! Instead of limiting the incoming request size (which usually causes trouble when doing file uploads (ask Vapor for example)), I'd suggest implementing streaming with back-pressure. That way the maximum amount of memory can remain constant and yet you can support arbitrarily large file uploads. Of course a user of Kitura should be able to control what's too much. So in some JSON API, maybe 10 MB might be a maximum but for another route that accepts file uploads there shouldn't be a maximum because files can be streamed straight to disk (possibly policing the quota for the logged in user of course).

[...]

PR #221, opened 2019-08-20
feat: Ability to limit request size and connection count 

RudraniW:
### **Description**
 
This PR allows a requestSizeLimit (maximum bytes for a request, including the header size) and connectionLimit (total number of concurrent connections) to be configured when registering a server.

Example usage:
`let serverConfig = HTTPServerConfiguration(requestSizeLimit: 1024, connectionLimit: 1024)`

If you do not specify either of the parameters, or do not specify a`HTTPServerConfiguration` at all, then default values are used. The defaults are defined in `HTTPServerConfiguration` in Kitura-NIO here: [https: //github.com/RudraniW/Kitura-NIO/blob/requestSize/Sources/KituraNet/HTTP/HTTPServerConfiguration.swift](url)

### **Motivation and Context**
See: #206 and https://github.com/IBM-Swift/Kitura/issues/1384


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -660,3 +673,23 @@ enum KituraWebSocketUpgradeError: Error {
     // Unknown upgrade error
     case unknownUpgradeError
 }
+
+class Atomic<A> {
```
weissi: you could use [NIO's Atomic<T>](https://apple.github.io/swift-nio/docs/current/NIOConcurrencyHelpers/Classes/Atomic.html) which actually is atomic and doesn't require a lock. Just add `NIOConcurrencyHelpers` to `Package.swift` and `import NIOConcurrencyHelpers`

RudraniW:
I am closing this PR as I have implemented the other approach to limit request size and connection count in #222 .

Sources/KituraNet/HTTP/HTTPServerConfiguration.swift
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public struct HTTPServerConfiguration {
```
djones6: The naming and facilities on this struct need to be consistent between Kitura-net and Kitura-NIO.  In https://github.com/IBM-Swift/Kitura-net/pull/307 I named this `ServerOptions`.  I don't have a strong preference as to which name we go with, however you'll need to change this anyway as there is a facility for sending a custom response back to the client which would need replicating here.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -127,22 +129,35 @@ public class HTTPServer: Server {
 
     var quiescingHelper: ServerQuiescingHelper?
 
+    private var ctx: ChannelHandlerContext?
+    
+    /// server configuration
+    public var serverConfig: HTTPServerConfiguration
```
djones6: This needs to be consistent with Kitura-net:
```swift
public var options: ServerOptions
```

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -127,22 +129,35 @@ public class HTTPServer: Server {
...
      ### Usage Example: ###
      ````swift
-     let server = HTTPServer()
+     let config =HTTPServerConfiguration(requestSize: 1000, coonectionLimit: 100)
+     let server = HTTPServer(serverconfig: config)
      server.listen(on: 8080)
      ````
     */
-    public init() {
+    public init(serverConfig: HTTPServerConfiguration = .default) {
```
djones6: Kitura-net's initializer does not take a parameter.  Instead, `HTTPServer.options` has an initial value of the default options, and is set by the user after the server is created, if desired.
(See usage in https://github.com/IBM-Swift/Kitura/pull/1481)

Sources/KituraNet/HTTP/HTTPRequestHandler.swift
```diff
@@ -76,12 +77,42 @@ internal class HTTPRequestHandler: ChannelInboundHandler, RemovableChannelHandle
...
-
         switch request {
         case .head(let header):
+            serverRequest = HTTPServerRequest(channel: context.channel, requestHead: header, enableSSL: enableSSLVerification)
+            if let requestSizeLimit = server.options.requestSizeLimit,
+                let contentLength = header.headers["Content-Length"].first,
+                let contentLengthValue = Int(contentLength) {
+                if contentLengthValue > requestSizeLimit {
+                    do {
+                        if let (httpStatus, response) = server.options.requestSizeResponseGenerator(requestSizeLimit, "") {
```
djones6: You are passing an empty string here for the client's identity.  This is used by the default response generator to log a message indicating which client is being rejected.  For example, Kitura-net uses the remote IP address + port of the client.

Sources/KituraNet/HTTP/HTTPRequestHandler.swift
```diff
@@ -152,4 +199,20 @@ internal class HTTPRequestHandler: ChannelInboundHandler, RemovableChannelHandle
     func updateKeepAliveState() {
         keepAliveState.decrement()
     }
+
+    func channelInactive(context: ChannelHandlerContext, httpServer: HTTPServer) {
+        httpServer.connectionCount.sub(1)
+    }
+    
+    func getHeaderSize(of header: HTTPRequestHead) -> Int {
```
djones6: Could this function be removed now?

Tests/KituraNetTests/LargePayloadTests.swift
```diff
@@ -80,13 +80,15 @@ class LargePayloadTests: KituraNetTest {
 
         func handle(request: ServerRequest, response: ServerResponse) {
             if  request.method.uppercased() == "GET" {
+                print("hello world")
```
djones6: These print statements probably want to be removed?

Tests/KituraNetTests/KituraNetWebSocketUpgrade.swift
```diff
@@ -31,7 +31,7 @@ class KituraNetWebSocketUpgradeTest: KituraNetTest {
         ]
     }
     
-    var httpHandler: HTTPResponseHandler?
+    var httpHandler: ResponseHandler?
```
djones6: This rename doesn't seem to belong to the same set of changes

Tests/KituraNetTests/ConnectionLimitTests.swift
```diff
@@ -0,0 +1,93 @@
...
+        do {
+            try channel = bootstrap.connect(host: "localhost", port: self.port).wait()
+            let request = HTTPRequestHead(version: HTTPVersion(major: 1, minor: 1), method: .GET, uri: "/")
+            self.sendRequest(request: request, on: channel)
+        } catch _ {
+            XCTFail("Connection is not established.")
+        }
+    }
+
+func testConnectionLimit() {
```
djones6: Can you fix the indentation for the rest of this file? It's difficult to tell whether (for example) this function and the classes defined below are within the scope of `ConnectionLimitTests` or not.

Sources/KituraNet/HTTP/RequestsizeHandler.swift
```diff
@@ -0,0 +1 @@
+
```
djones6: This appears to be a new file with only whitespace in it, can it be removed?

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -309,15 +324,16 @@ public class HTTPServer: Server {
             }
             .childChannelInitializer { channel in
                 let httpHandler = HTTPRequestHandler(for: self)
-                let config: NIOHTTPServerUpgradeConfiguration = (upgraders: upgraders, completionHandler: { _ in
+                let config: HTTPUpgradeConfiguration = (upgraders: upgraders, completionHandler: { ctx in
+                    self.ctx = ctx
```
djones6: Does this new property `ctx` ever get read? I can't see reference to it elsewhere in this PR

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -309,15 +324,16 @@ public class HTTPServer: Server {
...
                     _ = channel.pipeline.removeHandler(httpHandler)
                 })
                 return channel.pipeline.configureHTTPServerPipeline(withServerUpgrade: config, withErrorHandling: true).flatMap {
                     if let nioSSLServerHandler = self.createNIOSSLServerHandler() {
                         _ = channel.pipeline.addHandler(nioSSLServerHandler, position: .first)
                     }
                     return channel.pipeline.addHandler(httpHandler)
-                }
+                    }
```
djones6: whitespace error

RudraniW:
@djones6 Thanks for your feedback.
I have made the changes you asked.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -127,22 +129,35 @@ public class HTTPServer: Server {
 
     var quiescingHelper: ServerQuiescingHelper?
 
+    private var ctx: ChannelHandlerContext?
```
djones6: Please remove this unused var

Tests/KituraNetTests/LargePayloadTests.swift
```diff
@@ -87,6 +87,7 @@ class LargePayloadTests: KituraNetTest {
         }
 
         func handleGet(request: ServerRequest, response: ServerResponse) {
+            print("handle Get")
```
djones6: Did you intend to remove this one too?  The only other change in this file is a whitespace change, so I'd suggest just reverting this file.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -127,22 +129,35 @@ public class HTTPServer: Server {
...
+    
+    /// server configuration
+    public var options: ServerOptions = ServerOptions()
+
+    //counter for no of connections
+    var connectionCount = Atomic(value: 0)
+
+    // The data to be written as a part of the response.
+    //private var buffer: ByteBuffer
```
djones6: Remove this commented out code

PR #225, opened 2019-10-07
Add api to access eventloopGroup in Kitura-NIO

harish1992:
EventLoopGroup `eventloopGroup` was inacessible in previous versions of
Kitura-NIO. This PR makes `eventLoopGroup` a public api and also provides a way
to start HTTPserver with user created EventLoopGroup.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,8 +26,8 @@ let package = Package(
             targets: ["KituraNet"])
     ],
     dependencies: [
-        // FIXME: remove version constraint once IBM-Swift/Kitura-NIO#225 is merged
-        .package(url: "https://github.com/apple/swift-nio.git", "2.3.0"..."2.8.0"),
+        // Dependencies declare other packages that this package depends on.
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.8.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.0.0"),
         .package(url: "https://github.com/IBM-Swift/BlueSSLService.git", from: "1.0.0"),
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -137,6 +137,11 @@ public class HTTPServer: Server {
 
     /// The EventLoopGroup used by this HTTPServer. This property may be assigned
     /// once and once only, by calling `setEventLoopGroup(value:)` before `listen()` is called.
+    /// Server runs on `eventLoopGroup` which it is initialized to i.e. when user explicitly provides `eventLoopGroup` for server,
+    /// public variable `eventLoopGroup` will  return value stored private variable `_eventLoopGroup` when `ServerBootstrap` is called in `listen()`
+    /// making the server run of userdefined EventLoopGroup. If the `setEventLoopGroup(value:)` is not called, `nil` in variable `_eventLoopGroup` forces
+    /// Server to run in `globalELG` since value of `eventLoopGroup` in `ServerBootstrap(group: eventLoopGroup)` gets initialzed to value `globalELG`
+    /// if `setEventLoopGroup(value:)` is not called before `listen()`
     public var eventLoopGroup: EventLoopGroup {
         if let value = self._eventLoopGroup { return value }
         let value = globalELG
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -715,8 +715,8 @@ enum KituraWebSocketUpgradeError: Error {
     case unknownUpgradeError
 }
 
-// HTTP server errors
-public struct HTTPServerError: Error,Equatable {
+/// Errors thrown by HTTPServer
+public struct HTTPServerError: Error, Equatable {
 
     internal enum HTTPServerErrorType: Error {
         case eventLoopGroupAlreadyInitialized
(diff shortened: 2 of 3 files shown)


Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -145,6 +145,10 @@ public class HTTPServer: Server {
 #endif
     }
 
+    public init(eventLoopGroup: EventLoopGroup) {
```
djones6: It's going to be difficult to modify Kitura in a non-breaking way to call this new API.

Rather than requiring the ELG to be passed in at initialization, perhaps you could make the `eventLoopGroup` property lazily initialized?  Leave `init()` with no parameters, and I think something like this will work:

```swift
// A lazily initialized EventLoopGroup, accessed via `eventLoopGroup`
private var _eventLoopGroup: EventLoopGroup?

/// The EventLoopGroup used by this HTTPServer. This property may be assigned
/// once and once only, by calling `setEventLoopGroup(value:)` before `listen()` is called.
public var eventLoopGroup: EventLoopGroup {
    if let value = self._eventLoopGroup { return value }
    #if os(Linux)
        let numberOfCores = Int(linux_sched_getaffinity())
        let value = MultiThreadedEventLoopGroup(numberOfThreads: numberOfCores > 0 ? numberOfCores : System.coreCount)
    #else
        let value = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
    #endif
    self._eventLoopGroup = value
    return value
}

/// Sets the EventLoopGroup to be used by this HTTPServer. This may be called once
/// and once only, and must be called prior to `listen()`.
/// - Throws: If the EventLoopGroup has already been assigned.
public func setEventLoopGroup(_ value: EventLoopGroup) throws {
    guard _eventLoopGroup == nil else {
        throw HTTPServerError.eventLoopGroupAlreadyInitialized
    }
    _eventLoopGroup = value
}
```

Package.swift
```diff
@@ -28,7 +28,7 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.3.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", .branch("master")),
```
djones6: What release of `swift-nio` are we waiting for here?

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -83,8 +91,8 @@ public class HTTPServer: Server {
      ````
     */
     public private(set) var state: ServerState {
-        get {
-            return self.syncQ.sync {
+	get {
+		    return self.syncQ.sync {
```
djones6: Please can you fix this whitespace (tabs -> spaces)

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -684,3 +702,7 @@ enum KituraWebSocketUpgradeError: Error {
     // Unknown upgrade error
     case unknownUpgradeError
 }
+
+public enum HTTPServerError: Error {
```
djones6: Please can you make this an open enum (struct with static cases) as it's likely to be extended in the future

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -141,6 +141,10 @@ public class HTTPServer: Server {
     /// making the server run of userdefined EventLoopGroup. If the `setEventLoopGroup(value:)` is not called, `nil` in variable `_eventLoopGroup` forces
     /// Server to run in `globalELG` since value of `eventLoopGroup` in `ServerBootstrap(group: eventLoopGroup)` gets initialzed to value `globalELG`
     /// if `setEventLoopGroup(value:)` is not called before `listen()`
+    /// If you are using Kitura-NIO and need to access EventLoopGroup that Kitura uses, you can do so like this:
+    ///
+    ///         ```let eventLoopGroup = server.eventLoopGroup```
```
djones6: \`\`\` is a multi-line comment, you need line breaks here

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -684,3 +714,19 @@ enum KituraWebSocketUpgradeError: Error {
     // Unknown upgrade error
     case unknownUpgradeError
 }
+
+// HTTP server errors
```
pushkarnk: Documentation comment needed here.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -684,3 +714,19 @@ enum KituraWebSocketUpgradeError: Error {
...
+public struct HTTPServerError: Error,Equatable {
```
pushkarnk: `Error,Equatable` could be `Error, Equatable`

Tests/KituraNetTests/RegressionTests.swift
```diff
@@ -190,6 +192,87 @@ class RegressionTests: KituraNetTest {
...
+            defer {
+                server.stop()
+            }
+            do {
+                try server.listen(on: serverPort)
+            } catch {
+               XCTFail("Unable to start the server \(error)")
+            }
+            var goodClient = try GoodClient()
+            /// Connect a 'good' (SSL enabled) client to the server
```
pushkarnk: ` ///` not needed here

Tests/KituraNetTests/RegressionTests.swift
```diff
@@ -190,6 +192,87 @@ class RegressionTests: KituraNetTest {
...
+            defer {
+                server2.stop()
+            }
+            do {
+                try server2.listen(on: serverPort2)
+            } catch {
+                XCTFail("Unable to start the server \(error)")
+            }
+            var goodClient2 = try GoodClient()
+            /// Connect a 'good' (SSL enabled) client to the server
```
pushkarnk: ` ///` not needed here

Tests/KituraNetTests/RegressionTests.swift
```diff
@@ -190,6 +192,87 @@ class RegressionTests: KituraNetTest {
...
+        } catch {
+            XCTFail("Error: \(error)")
+        }
+        waitForExpectations(timeout: 10)
+    }
+
+    // Tests eventLoopGroup initialization in server after starting the server
+    // If server `setEventLoopGroup` is called after function `listen()` server should throw
+    // error HTTPServerError.eventLoopGroupAlreadyInitialized
+    func testFailEventLoopGroupReinitialization() {
```
pushkarnk: Please fix the indentation in this function

PR #232, opened 2019-11-17
move CLinuxHelpers in Linux target

makleso6:
I am using Kitura-NIO on iOS, and it no need to use CLinuxHelpers on Mac or iOS

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,29 @@
 
 import PackageDescription
 
+#if os(Linux)
+var targets: [PackageDescription.Target] = [
+    .target(
+        name: "CLinuxHelpers",
+        dependencies: []),
+    .target(
+        name: "KituraNet",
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -18,6 +18,29 @@
 
 import PackageDescription
 
+#if os(Linux)
```
weissi: this change will break cross-compilation. Generally in `Package.swift` files there should be no `#if os(...)`/`#if arch(...)` etc. I think the right change would be to make `CLinuxHelpers` compile on all OSes by using `#if __linux` etc in the .c/.h files similar to what SwiftNIO does for its `CNIOLinux` module: https://github.com/apple/swift-nio/tree/master/Sources/CNIOLinux

makleso6: Maybe you are right)
I rollback changes

makleso6:
@djones6 can you update snapshot in this repo?

PR #234, opened 2020-01-22
Fix warnings caused by API deprecations

pushkarnk:


Sources/KituraNet/HTTP/HTTPRequestHandler.swift
```diff
@@ -93,7 +93,7 @@ internal class HTTPRequestHandler: ChannelInboundHandler, RemovableChannelHandle
                     } catch {
                         Log.error("Failed to send error response")
                     }
-                    context.close()
+                    _ = context.close()
```
weissi: if you use `context.close(promise: nil)`, then that's more efficient and doesn't return anything so you don't need the `_ =`

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -347,7 +347,7 @@ public class HTTPServer: Server {
         }
 
         let bootstrap = ServerBootstrap(group: eventLoopGroup)
-            .serverChannelOption(ChannelOptions.backlog, value: BacklogOption.Value(self.maxPendingConnections))
+            .serverChannelOption(ChannelOptions.backlog, value: ChannelOptions.Types.BacklogOption.Value(self.maxPendingConnections))
```
weissi: use `ChannelOptions.backlog`, that works in all NIO versions. The `Types` are more or less internal

weissi (commented):
this looks good, just a few comments

Sources/KituraNet/HTTP/HTTPRequestHandler.swift
```diff
@@ -93,7 +93,7 @@ internal class HTTPRequestHandler: ChannelInboundHandler, RemovableChannelHandle
                     } catch {
```
weissi: you also need to bump the minimum NIO version in `Package.swift` to `from: "2.12.0"` because you're using new stuff from 2.12

pushkarnk:
Thanks for reviewing @weissi ! I've addressed your comments.

PR #239, opened 2020-04-01
modification in ChannelInactive of HTTPRequestHandler.

RudraniW:
This PR makes corrections in ChannelInactive of HTTPRequestHandler to decrement connectionCount. This fixes issue https://github.com/IBM-Swift/Kitura/issues/1501


mbarnach:
Nice catch! Indeed it fixes IBM-Swift/Kitura#1501!
I've cherry-picked the solution in PR#238 to try to address some more issues (crashes if keep alive is set to anything else than unlimited, etc.)
I think the CI failure is not relevant, can we re-run it and merge?

PR #244, opened 2020-09-28
Limit connection earlier

mbarnach:
As pointed by @davidde94 (https://github.com/Kitura/Kitura-NIO/issues/242), the connection should be closed earlier than on the end of the request.
This avoid dumping massive amount of data before checking the connection limit.


Davidde94 (approved):
LGTM

PR #248, opened 2021-03-14
Misc fix test

dannys42:


dannys42:
Woohoo!  It's passing!  Hopefully once merged I'll be able to get the Kitura tests to pass too.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -535,11 +535,12 @@ public class ClientRequest {
      */
     public func end(close: Bool = false) {
         closeConnection = close
-
+        let percentEncodedURL = URL(string: self.percentEncodedURL)!
```
mbarnach, line 538: Are you sure about the force unwrapping? It looks dangerously un-recoverable.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -535,11 +535,12 @@ public class ClientRequest {
...
+        
         var isHTTPS = false
 
         let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-        if (URL(string: percentEncodedURL)?.scheme)! == "https" {
+        if (percentEncodedURL.scheme)! == "https" {
```
mbarnach, line 543: Same here, even if it was already forced before.

.kitura-test.sh
```diff
@@ -10,11 +12,16 @@ if [ $SWIFT_TEST_STATUS -ne 0 ]; then
   return $SWIFT_TEST_STATUS
 fi
 
+# For now, short-circuit kitura tests until those are stabalized.
+return 0
```
mbarnach, line 16: I've create issue #249 to remember that...

mbarnach (changes requested):
I would feel better if we could avoid the force unwrapping, otherwise, good job!

Sources/KituraNet/ClientRequest.swift
dannys42: Yeah, I don't know how you'd recover from a missing scheme?  But I'll take another look... maybe there's a better way.

.kitura-test.sh
adam-rocska: I'd suggest to add a `// TODO: ` statement as well. Most IDEs collect and aggregate them. I for example use CLion, and this way I won't lose track of unkept promises.

.travis.yml
```diff
@@ -15,17 +15,23 @@ matrix:
       dist: xenial
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=swift:5.0.3-xenial SWIFT_SNAPSHOT=5.0.3 DOCKER_PRIVILEGED=true SWIFT_TEST_ARGS="--parallel"
+      env: DOCKER_IMAGE=docker.kitura.net/kitura/swift-ci:5.0.3 SWIFT_SNAPSHOT=5.0.3 DOCKER_PRIVILEGED=true SWIFT_TEST_ARGS="--parallel"
```
adam-rocska, line 18: wow

Sources/KituraNet/ClientRequest.swift
adam-rocska: If we'd have a huge concrete wall of tests, I'd be happy with it, but in its current state my lack of understanding and the code I see scares me too. 

But I don't have any better approaches right now either.

Personally I think force unwrapping is fine if there are multiple good quality and self validating unit tests which reason about the implementation and is convincing of its safety.

Otherwise I'd take the route of being super-defensive; and log warnings & errors whenever the sad path kicks in.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -550,8 +551,20 @@ public class ClientRequest {
...
+        if let port = percentEncodedURL.port {
+            portNumber = port
+        } else {
+            if percentEncodedURL.scheme == "https" {
+                portNumber = 443
+            } else if percentEncodedURL.scheme == "http" {
+                portNumber = 80
+            } else {
+                portNumber = 8080
+                Log.error("Unknown port for url: \(url)  Assuming 8080")
```
adam-rocska, line 565: Well, technically it's not an error, since you have recovered in a transparent & seamless way.
I'd suggest implementing it as a warning instead of an error.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -550,8 +551,20 @@ public class ClientRequest {
...
+            }
+        }
         if self.headers["Host"] == nil {
            let isNotDefaultPort = (portNumber != 443 && portNumber != 80) //Check whether port is not 443/80
```
adam-rocska, line 569: This whole play of port literals gave me an idea:

We could introduce a small harmless enum in one of the Kitura base modules which would encapsulate the typical ports and give names to them?

The bigger benefit is that it could open the door for future additional features to port handling without change risks.

@dannys42 && @mbarnach what do you think?

mbarnach: I like the idea: it makes it more obvious (like which port is the default for Kitura), but also keep tracks of standard and custom ports in a robust fashion.

dannys42: Hmm.. you mean an enum equivalent to the /etc/services file?  Might be interesting, but not sure it belongs in Kitura since Kitura really only deals with web traffic?

.travis.yml
dannys42: ?

Sources/KituraNet/ClientRequest.swift
mbarnach: The way I understand it is: instead of 80, you have something like `.http` and instead of 443, `.https`. For 8080, it would be `.default` (names are poorly chosen!). It could also be extended to use `.localhost` and `.domain(let custom)` for the hostnames. Feature wise, the same, but better to add semantic and group the value in a single place. As part of the port enum, the `var isDefaultPort: Bool { self == .http || self == .https }` could be introduced.

Sources/KituraNet/ClientRequest.swift
dannys42: I'm also a bit leery about it, but I was opting for minimal changes.  I think it will take quite a bit more restructuring to better handle these situations.

Sources/KituraNet/ClientRequest.swift
dannys42: My goal right now is just to get tests consistently passing.  So if you can agree that I at least haven't introduced any new problems, can we save the larger restructuring until later?

dannys42: Curiously the existing code does not use URL() as it's base way of storing URLs.  I think one reason for some of the complexity in the code is the way Kitura handles http/https vs Unix domain sockets.  If we're keeping this functionality (which I think is nice to have), it might make some of the code easier to follow one of these (non-standard) conventions for naming Unix domain sockets:  https://github.com/whatwg/url/issues/577

Sources/KituraNet/ClientRequest.swift
dannys42: That's true... but at the same time I don't think having a default 8080 port makes any sense.  I kept it there to avoid introducing more error paths and since the previous code had it.  You think it should be a warning instead of an error?

mbarnach:
@dannys42 Shall we create the tickets to keep track of those and merge the PR?

mbarnach (approved):

dannys42:
Created an issue to keep track of the various topics discussed here: https://github.com/Kitura/Kitura-NIO/issues/250

dannys42:
I'm not sure why, but with Xcode 10.3, swift-nio-ssl doesn't build properly.  Looking at the source, it should be requiring macOS 10.15, but Travis-CI's Xcode 10.3 image is clearly running macOS 10.14.  So I'm not sure why there was an issue.  I'm just going to deprecate Xcode 10.3.  We do see that Swift 5.0.1 still builds fine on Linux so I think this is probably fine.

PR #251, opened 2021-04-05
Fix numerous swift warnings

dannys42:


dannys42:
Not sure why the tests aren't passing on the main branch, but maybe fixing some of these warnings will help.

mbarnach (approved):
Looks good to me! 👍

Sources/KituraNet/BufferList.swift
```diff
@@ -147,7 +147,12 @@ public class BufferList {
 
     */
     public func fill(array: inout [UInt8]) -> Int {
-        return fill(buffer: UnsafeMutablePointer(mutating: array), length: array.count)
+        var filledBytes = 0
+        let arrayCount = array.count
+        array.withUnsafeMutableBufferPointer({ bufferPointer in
+            filledBytes = fill(buffer: bufferPointer.baseAddress!, length: arrayCount)
```
weissi, line 153:
```suggestion
            filledBytes = fill(buffer: bufferPointer.baseAddress!, length: bufferPointer.count)
```

you shouldn't really refer to `array` in here

Sources/KituraNet/BufferList.swift
```diff
@@ -147,7 +147,12 @@ public class BufferList {
 
     */
     public func fill(array: inout [UInt8]) -> Int {
-        return fill(buffer: UnsafeMutablePointer(mutating: array), length: array.count)
+        var filledBytes = 0
+        let arrayCount = array.count
```
weissi, line 151:
```suggestion
```

Sources/KituraNet/HTTP/HTTPRequestHandler.swift
```diff
@@ -93,7 +93,7 @@ internal class HTTPRequestHandler: ChannelInboundHandler, RemovableChannelHandle
                     } catch {
                         Log.error("Failed to send error response")
                     }
-                    context.close()
+                    _ = context.close()
```
weissi, line 96:
```suggestion
                    context.close(promise: nil)
```

is the correct (and faster (saves an allocation)) way to say "I don't care about the result"

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -347,7 +343,7 @@ public class HTTPServer: Server {
         }
 
         let bootstrap = ServerBootstrap(group: eventLoopGroup)
-            .serverChannelOption(ChannelOptions.backlog, value: BacklogOption.Value(self.maxPendingConnections))
+            .serverChannelOption(ChannelOptions.backlog, value: ChannelOptions.Types.BacklogOption.Value(self.maxPendingConnections))
```
weissi, line 346:
```suggestion
            .serverChannelOption(ChannelOptions.backlog, value: self.maxPendingConnections)
```

mbarnach:
@weissi Thanks for the good tips!

Sources/KituraNet/HTTP/HTTPRequestHandler.swift
dannys42: Thanks!

Sources/KituraNet/BufferList.swift
dannys42: Hi @weissi thanks for helping to review this PR.  Can you suggest a better way to accomplish this?  Based on the description in the comments the goal of this function is to fill `array` with the contents of the bufferList so I must refer to `array` in some way.  Unfortunately with the new `withUnsafeMutableBufferPointer` form, I cannot refer to `array.count` directly in the block, hence the local variable to keep track of the length of array.

Sources/KituraNet/HTTP/HTTPServer.swift
dannys42: Unfortunately `maxPendingConnections` is an Int and "value" expects a type of `ChannelOptions.Types.BacklogOption.Value` so I cannot make this change.

Sources/KituraNet/BufferList.swift
dannys42: (see comment below)

Sources/KituraNet/HTTP/HTTPServer.swift
weissi: ah, okay, you could write `CInt(self.maxPendingConnections)` or `Int32(self.maxPendingConnections)` or just `.init(self.maxPendingConnections)` if you wanted but I get your point

Sources/KituraNet/BufferList.swift
weissi: @dannys42 but `bufferPointer.count` will have the same value as `array.count` and you should be using that one.

However, two observations:
- It should never be necessary to use `*Unsafe*` stuff in Kitura-NIO (if it is, please file a NIO bug). And in this case, the easiest way to write the code you have there would probably be:

```swift
    public func fill(array: inout [UInt8]) -> Int {
        // I don't know why this API doesn't take all the available bytes, but let's stick to its contract.
        let bytesToFill = min(array.count, self.byteBuffer.readableBytes)

        // Get a `ByteBufferView` (a `Collection`) of all readable bytes.
        let readableByteView = self.byteBuffer.readableBytesView

        // We want to overwrite the first `bytesToFill` bytes with the first `bytesToFill` bytes from the `ByteBufferView`.
        array.replaceSubrange(0 ..< bytesToFill,
                              with: readableByteView.prefix(bytesToFill))

        // And then move the reader index to indicate that we consumed the bytes.
        self.byteBuffer.moveReaderIndex(forwardBy: bytesToFill)

        return bytesToFill
    }
```

- is anybody actually using these APIs?

Sources/KituraNet/BufferList.swift
```diff
@@ -147,7 +147,12 @@ public class BufferList {
...
+        array.withUnsafeMutableBufferPointer({ bufferPointer in
+            filledBytes = fill(buffer: bufferPointer.baseAddress!, length: arrayCount)
+        })
+        return filledBytes
```
weissi, line 155: btw, for the `fill` into an `UnsafeMutablePointer` (which should be deprecated), I'd suggest

```swift
    public func fill(buffer: UnsafeMutablePointer<UInt8>, length: Int) -> Int {
        // NOTE: This API should be deprecated and expressed as either `UnsafeMutableBufferPointer<UInt8>` or
        // preferrable `UnsafeMutableRawBufferPointer` or even better: replaced with nothing.

        let bytesToFill = min(length, self.byteBuffer.readableBytes)

        // Get a `ByteBufferView` (a `Collection`) of all readable bytes.
        let readableByteView = self.byteBuffer.readableBytesView

        // We want to overwrite the first `bytesToFill` bytes with the first `bytesToFill` bytes from the `ByteBufferView`.
        _ = UnsafeMutableBufferPointer(start: buffer, count: bytesToFill)
            .initialize(from: readableByteView.prefix(bytesToFill))

        // And then move the reader index to indicate that we consumed the bytes.
        self.byteBuffer.moveReaderIndex(forwardBy: bytesToFill)

        return bytesToFill
    }
```

Sources/KituraNet/BufferList.swift
dannys42: @weissi thanks!  The Kitura-NIO package is attempting to be API compatible with KituraNet (which was created pre-NIO).  Right now I'm just trying to update the existing code to avoid the compiler warnings.  Since I'm not sure wasn't sure what the performance impact is of moving to pure swift functions, nor am I that familiar with NIO, I opted for making the smallest amount of changes.  But your changes definitely look clean, so I'm happy to incorporate them.

I'm not sure if there are other packages using this fill() API.  But KituraNet does appear to be using it for FastCGI support, which does not look like it's implemented yet in Kitura-NIO.  So it's possible it's not an issue at this time.  But I'll be updating KituraNet with similar changes in order to get rid of these warnings.

Sources/KituraNet/HTTP/HTTPServer.swift
dannys42: Oh I see.  I like the `.init()` approach.  Thanks for the tip!

Sources/KituraNet/BufferList.swift
dannys42: Thanks again!

Sources/KituraNet/BufferList.swift
dannys42: FYI, update to Kitura-net is in this [PR](https://github.com/Kitura/Kitura-net/pull/321)

weissi: Cool, sounds good! Regarding FastCGI: I would very much recommend to not support it (is there any point these days?) but if you wanted to support it, I'd transparently support it in Kitura-NIO.

dannys42:
@weissi Thanks again for your review.  If everything looks good to you, I'm going to merge.

PR #254, opened 2021-09-29
Support latests NIO code; Make Swift 5.2 the minimum supported version

dannys42:


mbarnach (approved):
I think 5.2 could be a problem for Raspberry PI arch, but it seems un-avoidable due to NIO.
All good otherwise!

